Issue #88, opened 2014-08-15
How to set connection timeout?


feiyao:
Is there anyway to set the timeout on the connection?

Thanks
Fei

cressie176:
Can be done via socket options

PR #118, opened 2014-11-28
Allow specifying connection timeout in socket options

rexxars:
Since `connect()` does not return the socket instance, there is currently no easy way to set the connection timeout. This PR adds a `timeout` option to the socket options passed to `connect()`, and automatically sets it on the socket when connecting. If a timeout is reached, it will trigger the connect callback with an error.

This should solve #88 .

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -101,6 +101,7 @@ function connect(url, socketOptions, openCallback) {
   var parts = URL.parse(url, true); // yes, parse the query string
   var protocol = parts.protocol;
   var noDelay = !!sockopts.noDelay;
+  var timeout = sockopts.timeout;
 
   var extraClientProperties = sockopts.clientProperties || {};
   var credentials = sockopts.credentials || credentialsFromUrl(parts);
@@ -116,6 +117,9 @@ function connect(url, socketOptions, openCallback) {
   function onConnect() {
     sockok = true;
     sock.setNoDelay(noDelay);
+    sock.setTimeout(timeout || 0, function() {
+        openCallback(new Error('connect ETIMEDOUT'));
+    });
     var c = new Connection(sock);
     c.open(fields, function(err, ok) {
       if (err === null) openCallback(null, c);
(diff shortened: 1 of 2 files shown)


jcrugzz:
I like this idea @rexxars but the timeout should be the default and not 0 if a timeout option is not specified. So i'd wrap that in an if clause instead.

rexxars:
I agree there should be a default timeout - however, the default as of now is no timeout. Introducing a default timeout might be considered to be a breaking change, so I'd want to OK that with the maintainer first.

Like I said, the default is 0, meaning no timeout. I could use an if clause, but in theory it should do pretty much the exact same as it does now.

jcrugzz:
@rexxars yea thats totally my fault. I was assuming HTTP here which does have a default timeout. My apologies, this does look good.

squaremo:
> I could use an if clause, but in theory it should do pretty much the exact same as it does now.

Supplying `0` to `Socket#setTimeout` means "clear any timeout", which is not what's intended. (http://nodejs.org/api/net.html#net_socket_settimeout_timeout_callback)
It is harmless as things stand, since the socket is created then and there, but it might be a minor timebomb.

rexxars:
I didn't bother doing it since the socket was created in place, but optimizing for change makes sense. Latest commit checks if the timeout specified in options is actually set before assigning the timeout handler.

squaremo:
Thanks @rexxars. I tweaked it slightly -- are you happy with 7ccd837334db22b56f87b0a0627d3fd1f2a52d1f ?

rexxars:
Yup, looks good!

zweifisch:
Seems that this does not affect the timeout value for `net.connect`(which is about 2 minutes on my machine), since `sock.setTimeout` is put inside the `onConnect` function. Is it intended?

squaremo:
> Seems that this does not affect the timeout value for net.connect(which is about 2 minutes on my machine), since sock.setTimeout is put inside the onConnect function. Is it intended?

That .. is a good point @zweifisch. Casting my mind back, I think perhaps it was only intended to cover a failure to complete the handshake. But failing to establish a connection seems like something it could also cover with the minor modification you imply.

What do you think, @rexxars?

rexxars:
I agree, this looks like an oversight. I think this fixed a specific problem I had, but moving it outside the onConnect method should cover both connection and handshake timeouts.

PR #119, opened 2014-12-02
fix connecting on uris with trailing slash and no vhost

hopkinsth:
This PR fixes the case where this would fail:

```js
amqp.connect('amqp://127.0.0.1:5672/')
```

but this would succeed:
```js
amqp.connect('amqp://127.0.0.1:5672');
```

The trailing slash in the first example becomes an empty vhost as determined in connect.js: https://github.com/hopkinsth/amqp.node/blob/a972f63edf9321c901cf016b3a6f642d72d54380/lib/connect.js#L53

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -47,7 +47,7 @@ var CLIENT_PROPERTIES = {
 function openFrames(parts, credentials, extraClientProperties) {
 
   var vhost = parts.pathname;
-  if (!vhost)
+  if (!vhost || vhost === '/')
     vhost = '/';
   else
     vhost = QS.unescape(vhost.substr(1));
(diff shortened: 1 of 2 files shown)


squaremo:
> Is it possible to send an empty vhost with RabbitMQ?

It is possible, but *really* inconvenient, to have a vhost with an empty name; for example, the management interface will let you create one, but you can't do anything with it (e.g., you can't give users permissions for it).

> RabbitMQ initially accepts the connection but closes it soon afterward... or something like that. I have not yet looked to see at what point it disconnects. I believe I'm using RabbitMQ 3.4.x locally. I can investigate a little more.

I'm going to close this PR to take it off the radar, but if you find what your problem was, please do log an issue or PR. Cheers!

PR #120, opened 2014-12-16
Bumping when-version to 3.6.2

jonfryd:
Work around v8 bug as described in https://github.com/cujojs/when/issues/345 by bumping when-version


squaremo:
Thanks for the PR! I didn't want the extra commits, so I did the same change on master at 63752b1d253e058ed3d6f38ca39ffecb74ae8e50.

PR #125, opened 2015-01-07
Add keep-alive and timeout options (thanks @rexxar)

jcrugzz:
When implementing keep-alive support I just pulled in the work from @rexxar on timeouts as i slightly refactored it to handle timeout events that could possibly happen after connect (which I have not seen but was going for the side of safety). @squaremo let me know your thoughts as I think having keep-alive support is a nice complement to the heartbeat functionality you have implemented.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -85,7 +85,7 @@ function credentialsFromUrl(parts) {
   if (parts.auth) {
     var auth = parts.auth.split(':');
     user = auth[0];
-    passwd = auth[1];    
+    passwd = auth[1];
   }
   return credentials.plain(user, passwd);
 }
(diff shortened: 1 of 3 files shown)


jcrugzz:
hmm its curious that 0.8.x fails here, looking into that.

lib/connect.js
```diff
@@ -101,6 +101,11 @@ function connect(url, socketOptions, openCallback) {
   var parts = URL.parse(url, true); // yes, parse the query string
   var protocol = parts.protocol;
   var noDelay = !!sockopts.noDelay;
+  // We should default to true as this nicely complements the heartbeat
+  var keepAlive = !!sockopts.keepAlive || true;
```
rexxars: This doesn't actually do what you want... If I specify ```{ keepAlive: false }```, this will basically say ``` var keepAlive = false || true;```, which obviously is always true. So you're basically always enabling keepAlive here.

jcrugzz: ah good catch, thanks!

squaremo:
To be honest I'm not sure if using TCP keep-alive is appropriate here or not. It might be a better way (than using AMQP's heartbeats) of keeping load-balanced and NAT connections open. Is that the idea, @jcrugzz ?

jcrugzz:
@squaremo the idea was to just be complementary to the heartbeats as they are both designed to keep the TCP socket open and detect when something has been broken.

PR #146, opened 2015-03-24
node: ">=0.8 <0.13 || ^1"

OronNadiv:
@squaremo - Here at Lanetix we are using amqplib with iojs v1 - v1.6.2 and it works well.
Since all iojs-v1.x.x should be backward compatible, there is no reason to specify a specific iojs version in the package.json file.
This change will remove the warning we see when we install amqplib using npm with the latest versions of iojs (ex: 1.6.2).

Thanks,
-Oron


michaelklishin:
:+1:

PR #148, opened 2015-04-16
[Docs] Channel.publish returns a boolean

salzhrani:


squaremo:
Thank you for this. The headings are meant to indicate which API the method belongs to, rather than the return type. Maybe they should be e.g.,

## connect
### Promises API
#### `connect([url, [socketOptions]]) -> Promise`

to make this clearer. What do you think?

salzhrani:
I agree.

PR #159, opened 2015-05-26
add support for passing connection parameters as an object

ben-page:
The built in url.parse() does not properly handle encoded variables in the username or password. Consequently, some username and/or passwords can not be based to amqplib as a URL. For reference, see: https://github.com/joyent/node/issues/25353

So, I updated connect() to support passing connection parameters as an object:
```javascript
var url = {
    protocol: 'amqp', //amqp or amqps
    username: 'someone',
    password: 'secret',
    hostname: '127.0.0.1',
    port: 5672,
    vhost: 'name'
}
connect([url, [socketOptions]], function(err, conn) {...})
```

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -108,12 +108,17 @@ function connect(url, socketOptions, openCallback) {
 	protocol = (url.protocol || 'amqp') + ':';
 	sockopts.host = url.hostname;
 	sockopts.port = url.port || ((protocol === 'amqp:') ? 5672 : 5671);
+	
+	var user, pass;	
     if (!url.username) {
-      url.username = 'guest';
-      url.password = url.password || 'guest';
-    }
+      user = 'guest';
+      pass = url.password || 'guest';
+    } else {
+      user = url.username;
+      pass = url.password;
+	}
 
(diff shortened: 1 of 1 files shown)


lib/connect.js
```diff
@@ -107,12 +102,26 @@ function connect(url, socketOptions, openCallback) {
...
   var extraClientProperties = sockopts.clientProperties || {};
-  var credentials = sockopts.credentials || credentialsFromUrl(parts);
+  
+  var protocol, fields;
+  if (typeof url === 'object') {
+	protocol = (url.protocol || 'amqp') + ':';
+	sockopts.host = url.hostname;
+	sockopts.port = url.port || ((protocol === 'amqp:') ? 5672 : 5671);
+    if (!url.username) {
+      url.username = 'guest';
```
squaremo: If `url` is mutated, perhaps it's best to clone it. This way the user won't be surprised if the value is used elsewhere.

ben-page:
@squaremo I agree with your comment. I updated accordingly.

squaremo:
Thank you! Sorry I took a while to get to this.

PR #164, opened 2015-06-23
Refactor example tutorial (send)

RebootJeff:


RebootJeff:
- Remove nested promise chains and use promise idioms
- Reduce amount of inlining for clarity

squaremo:
Hello, thank you for this. I'm not sure I agree that verbosity makes better examples -- it can introduce more noise. I do appreciate that one could see the composition at the bottom as much neater than nested callbacks `})})})})});`. I am a bit undecided about this change. Perhaps if I try doing the same thing to one of the more complicated examples I will be able to judge the value of this style.

It makes you realise how barely adequate JavaScript is for functional-style programming doesn't it -- a bit of partial application would help a lot, among other things.

`tap()` is unclear to me -- the name, rather than what it does -- is it a when-specific method?

RebootJeff:
Admittedly, I didn't check the tests, so it looks like this PR is dead in the water anyway (the "connect" tests are failing?). My bad!

**Functional Programming**
It's funny that you mention partial application because I've used currying from Ramda.js to clean up my promise chains when using amqp.node. JavaScript can be much more FP-friendly when you add the Ramda library.

**Tap Method**
I don't think `.tap()` is part of the Promises/A+ specification, but it's a popular helper in promise libraries like When, Q, and Bluebird. I believe most JS promise libraries have it by now.

squaremo:
It looks like a spurious error to me -- I've restarted the travis job

examples/tutorials/send.js
```diff
@@ -1,19 +1,41 @@
...
+  function closeConnection() {
+    if(connection) {
+      connection.close();
+    }
+  }
+  
+  return amqp.connect(url)
+    .then(createChannel)
+    .tap(setupQueue)
+    .tap(sendMsgToQueue)
```
jimmywarting, line 35: don't like seeing `tap` when native async/await becoming the new norm

RebootJeff: Keep in mind this PR was created 6 years ago. `async/await` wasn't the norm back then 😛

cressie176:
Sorry this took 6+ years to resolve @RebootJeff . amqplib has switched to native promises so tap is no longer available. Instead I've updated the examples for async/await, although think your approach is still more readable

PR #168, opened 2015-06-29
Remove unnecessary semicolon

enaqx:


squaremo:
Good spot! I think a PR I just merged might have covered this -- fancy giving it a quick rebase?

PR #169, opened 2015-06-30
handle net.connect() timeout

zweifisch:
handle `net.connect()` timeout as discussed in #118


zweifisch:
ping

squaremo:
Thanks! And sorry to take a while over this.

PR #170, opened 2015-07-06
add a new error on net or tls connect timeout

RoCat:
I don't know if your error throwing is still used with my changes, i don't think so. But without it, if I tryed to connect on an unused IP, I just ran to infinite async wait, no error was called.


squaremo:
Thanks! I believe #169 made the same useful change; so closing this one.

PR #178, opened 2015-08-13
Call EventEmitter within functions inheriting it

### LINKED ISSUE #177: EventEmitter not being called by functions inheriting
tl;dr: Why do we not set a domain on a connection

----------------------

Is there any reason not to call EventEmitter within things like Connection and Channel?

Lets take connection for example

We have this: 

`inherits(Connection, EventEmitter);`

Usually when using `inherits` you would also call the constructor

`EventEmitter.call( this )`

The reasoning for creating this issue is because if the constructor ( or `init` ) isn't called, when an event is emitted, the domain can't pick up on unhandled error events nor it can catch errors occurring in any other event. 

If there is a good reason for this not to be done then please explain, else I will make a pull request for this.

fabiancook:
Fixes #177

### DIFF
--- a/lib/callback_model.js
+++ b/lib/callback_model.js
@@ -15,6 +15,7 @@ var Args = require('./api_args');
 function CallbackModel(connection) {
   if (!(this instanceof CallbackModel))
     return new CallbackModel(connection);
+  EventEmitter.call( this );
   this.connection = connection;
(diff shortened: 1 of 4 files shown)


squaremo:
Yep that's a fair cop. Thanks!

PR #180, opened 2015-08-25
Support priority queue declarations

ebardes:
I'm working on project that benefits from using priority queues in RabbitMQ (available since version 3.5.x). I feel that others might benefit from this functionality too.

### DIFF
--- a/lib/api_args.js
+++ b/lib/api_args.js
@@ -57,6 +57,7 @@ Args.assertQueue = function(queue, options) {
   setIfDefined(argt, 'x-dead-letter-routing-key',
                options.deadLetterRoutingKey);
   setIfDefined(argt, 'x-max-length', options.maxLength);
+  setIfDefined(argt, 'x-max-priority', options.maxPriority);
 
   return {
     queue: queue,


michaelklishin:
@ebardes you already can use priority queues with any client that supports optional queue arguments. While I couldn't quickly find an example of how to do that with this client in either examples or tests, I'd be really surprised if amqp.node didn't provide a way to do that.

ebardes:
I couldn't find a way to declare a queue with the 'x-max-priority' flag. Perhaps I just missed it or that I'm not familiar enough with the nuances of amqp.node yet.

michaelklishin:
Ah, so there's known argument filtering going on.

ebardes:
Exactly. I just added another known argument.

nikosd23:
This is very important for us! When do you plan to perform a new release with this included?

michaelklishin:
@squaremo this PR looks good to me.

nikosd23:
Tested with RabbitMQ 3.3.4 broker and it seems to create the queues successfully, ignoring the x-max-priority option making the change backwards compatible

squaremo:
> Ah, so there's known argument filtering going on.

No there's not .. I see why you'd think that though. The second argument to `#assertQueue` is options, and an `x-*` argument would go in the _arguments_ table in the options:

    ch.assertQueue('prio', {arguments: {'x-max-priority': 10}});

whereas the "known options" go in the top level:

    ch.assertQueue('prio', {maxPriority: 10});

Anyway, yes this PR looks fine.

michaelklishin:
@nikosd23 all unknown `arguments` will be ignored. They are all optional.

PR #185, opened 2015-09-14
Return error when message is unroutable

### LINKED ISSUE #184: ConfirmChannel.publish should callback with an error for unroutable messages
When an unroutable message is published, a 'return' event is emitted, but the callback is called without an error.

KurtPattyn:
Closes #184

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -27,7 +27,14 @@ function Channel(connection) {
   this.lwm = 1; // the least, unconfirmed deliveryTag
   this.unconfirmed = []; // rolling window of delivery callbacks
   this.on('ack', this.handleConfirm.bind(this, function(cb) {
-    if (cb) cb(null);
+    if (cb) {
+      var err = null;
+      if (self._returnMessage && self._returnMessage.fields.replyCode === defs.constants.NO_ROUTE) {
+        err = new Error('message unroutable');
+        self._returnMessage = null;
+      }
+      cb(err);
+    }
   }));
   this.on('nack', this.handleConfirm.bind(this, function(cb) {
     if (cb) cb(new Error('message nacked'));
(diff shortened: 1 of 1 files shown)


squaremo:
Thank you for the PR!

Going on the description [here](http://www.rabbitmq.com/confirms.html), I don't think it's certain that the acknowledgement will arrive *exactly* after the returned message.

All that is guaranteed by RabbitMQ is that the acknowledgment will not arrive before the returned message. That seems to admit

 * an acknowledgment for another message arriving in the meantime
 * another returned message arriving in the meantime

In general, it is not possible to correlate returned messages with those sent, unless (say) an identifying value is put in the headers or body by the application. (Yes, this is an irritating mistake in AMQP.)

KurtPattyn:
@squaremo Hi Michael,

this is what I found in the discussion forum (from a Pivotal Software Engineer):
Question: `how can the client code match a returned message with the specific published message?`
Answer: `Either by waiting for the following basic.ack, or by matching message body 
and properties (e.g. taking checksums of both). I agree it would be helpful if returns 
had a delivery tag but publisher confirms is not a part of the original spec.`  

The link to this topic is:
https://groups.google.com/d/topic/rabbitmq-users/RRSFgt2A9Kw/discussion

So, I suppose that a basic.ack always follows the corresponding basic.return.

KurtPattyn:
It is indeed correct that it is not guaranteed that a basic.ack immediately follows upon a basic.return.
On the question **"when a  basic.return is received is it guaranteed that the next basic.ack refers to the same message?"** on the RabbitMQ forum, I got this answer:
**"No but it is typically the case and it is guaranteed that a basic.ack/nack for a message is sent after basic.return."**

Closing this pull request.

PR #186, opened 2015-09-14
Provided a tutorial for RabbitMq priority queue implementation

girirajsharma:
I have observed RabbitMq tutorials for different client's are quite consistent i.e., exactly same tutorials are available for all client's. Updating/adding tutorial of one of the client possibly requires porting the exact same tutorial for all other client's

Therefore, I would be happy to issue following PR's step wise as follow-up commits:
1. Porting the same tutorial for [Javascript (with Node and amqp-node)](https://github.com/squaremo/amqp.node/tree/master/examples/tutorials) (using promises/futures)
2. Porting the same tutorial for all [RabbitMq clients](https://github.com/rabbitmq/rabbitmq-tutorials) with a separate PR for each client in step-wise order.
3. If feasible and applicable, porting the tutorial demonstration over official [RabbitMq website](https://www.rabbitmq.com/getstarted.html) in step-wise order for supported client's as such the [website source](https://github.com/rabbitmq/rabbitmq-website/) is open.

Please correct me if above idea(s) doesn't really make sense or is(are) not applicable.

### DIFF
--- a/examples/tutorials/README.md
+++ b/examples/tutorials/README.md
@@ -80,6 +80,14 @@ and routing replies back to clients.
  * [rpc_server.js](rpc_server.js)
  * [rpc_client.js](rpc_client.js)
 
+## [Tutorial seven: Priority Queue][tute-seven]
+
+RabbitMQ has priority queue implementation in the core as of version 3.5.0.
+Using RabbitMQ as a priority queue, showing how to set maximum priority value supported by the queue and setting priority of each message pushed to the queue. In case of messages having equal value of priority, they are consumed in FIFO order.
+
+ * [send_priority_queue.js](callback_api/send_priority_queue.js)
+ * [receive_priority_queue.js](callback_api/receive_priority_queue.js)
+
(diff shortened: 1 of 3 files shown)


girirajsharma:
No idea why it fails on Node.js~0.9 and passes on all other versions.

squaremo:
Thanks for the PR :+1:

You should really put the idea of a priority queue tutorial to the RabbitMQ developers (maybe via https://groups.google.com/forum/#!forum/rabbitmq-users). They will probably want to have some text to go along with it, but maybe you could collaborate on that if they like the idea.

In the meantime, I welcome example code -- could you perhaps put send and receive in the same file, and move it to examples/?

girirajsharma:
Yeah, I will be pushing the idea on the suggested forum and will act accordingly. Infact, I am on it :)

It's fine to move the code to examples/ but i din't get putting send and receive in the same file. Isn't it better to keep them separate like all other tutorials.

cressie176:
Too old

PR #197, opened 2015-10-16
Run tests with docker

christiaan:
Added documentation on how to use docker to run the testsuite


squaremo:
Good idea, thanks! Way better than using dev.rabbitmq.com.

PR #198, opened 2015-10-16
in tutorials note that errors should be handled

christiaan:
SendToQueue and Publish can return false when the outgoing buffer is full.
The only place this is mentioned is in the API reference the tutorials do
not give the slightest hint

### DIFF
--- a/examples/tutorials/callback_api/emit_log.js
+++ b/examples/tutorials/callback_api/emit_log.js
@@ -18,6 +18,8 @@ function on_connect(err, conn) {
     var msg = process.argv.slice(2).join(' ') ||
       'info: Hello World!';
     ch.publish(ex, '', new Buffer(msg));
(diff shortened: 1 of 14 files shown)


squaremo:
That's a good point! Putting a comment in each tutorial is perhaps overkill though. How about a comment in the first tutorial, and I'll write a section on flow control for the top of the API docs, so it's not hidden down by `#publish` -- what do you think?

christiaan:
For me it does not matter much any more I just wish it was more obvious when I started with the library.

Of course you have to read the documentation properly when using a library. But I think most people copy paste an example as a starting point. So if the obvious example everyone copies as their starting point includes pointers on the return value that might be enough.

squaremo:
I've put a note in the first tutorial, and a section in the introductory matter of the API reference. Thanks for pointing this out to me, and taking time to submit the PR.

Issue #200, opened 2015-10-21
channel.cancel doesn't emit null message on channel.consume


ierceg:
The documentation states:

> If the consumer is cancelled by RabbitMQ, the message callback will be invoked with `null`.

However, this doesn't happen when `channel.cancel` is invoked manually. Is this the correct behavior? `null` is emitted when a temporary queue is deleted (once its channel is cancelled) but not for persistent queues.

I tried fixing it by emitting `cancel` event manually and that works but then `cancel consumer` test fails. I've been looking at it and I'm not sure I understand it. It seems to cancel the channel, then send to queue over the same channel and then checks that the message has indeed landed in the queue. But should any new message be sent to he queue if the channel was cancelled? Here's the relevant code:

```js
    return doAll(
      ch.cancel(ctag).then(function() {
        ch.sendToQueue(q, new Buffer('bar'));
      }),
      // but check a message did arrive in the queue
      waitForMessages(q))
      .then(function() {
        ch.get(q, {noAck:true})
          .then(function(m) {
            // I'm going to reject it, because I flip succeed/fail
            // just below
            if (m.content.toString() === 'bar')
              recv1.reject();
          });
        return expectFail(recv1.promise);
        // i.e., fail on delivery, succeed on get-ok
      });
```

ierceg:
PR #201 is supposed to fix this issue by handling `BasicCancelOk` message the same way `BasicCancel` is handled now (emitting `cancel`). I have also changed the previous "cancel consumer" test to be more like "deleted queue" test and check for `null` message in `consume` rather than try to send a message after channel has been cancelled.

squaremo:
> But should any new message be sent to he queue if the channel was cancelled?

It's a consumer that is being cancelled; the idea is that the message sent after the cancel won't be forwarded to a consumer, and so will be visible in the queue.

You have a point about the consumer being given `null` only if the server cancels -- that's inconsistent. What are the consequences in practice?

ierceg:
Aha, I got it now - it's the consumer getting cancelled, not channel getting closed. Alright, when I get some time I'll revisit the test, see how to rewrite it.

The consequence in practice is that there is no way to cancel `.consume` so it just hangs there forever. In any case I have moved away from `.consume` implementation and I now use `.get` so it's no longer important to me. But it seems like it should be consistent with the documentation.

PR #201, opened 2015-10-22
Add handler for BasicCancelOk message to fix no null message on cancelled consume

ierceg:
Fixes issue #200


ierceg:
The CI passes now.

PR #204, opened 2015-11-17
Split open frame options

csabapalfi:
Open frame options are now clearly separated. Makes it easier to read and we'll only send the relevant frame options during the handshake.

(While debugging a connection issue it caught me that I can't properly see frame options because of not-enumerable properties created by `Object.create`. Then I realised all frame options are sent all the time.)


csabapalfi:
Just realised that the encode methods defs.js already pick the correct values for each frame.

PR #205, opened 2015-11-17
Keep starting slash in vhost name for ConnectionOpen frame

csabapalfi:
I keep getting connection closed on ConnectionOpen while setting up an amqps connection and keeping the trailing slash fixes it. ([postwait/node-amqp](https://github.com/postwait/node-amqp) does the same thing, that's where I got the idea from)


csabapalfi:
Actually this can be worked around by just calling connect with the URL as an object and specifying a vhost with a starting slash there.

PR #206, opened 2015-11-20
Add v5 to tests

forstermatth:


squaremo:
Hello, thank you for the PR. I've added 5.5 to the Makefile and Travis tests now -- do you think it's worth me testing 5.1 and 5.2 too, or will people pretty much move straight to 5.5 (and its successors)?

forstermatth:
My inclination would be to stay with the most up to date unstable branch. Development should be focused on moving forwards with v5, while still testing and looking for stability in v4. 

It doesn't hurt to add the different versions for clarities sake though, if I was looking at a repo I'd definitely see that as a positive. 

Just my thoughts.

PR #219, opened 2016-01-18
test latest node. See nvm "node" alias

revington:
Travis uses nvm to install node. By adding the "node" option to `node_js` configuration you enforce your test to run on latest node version. 

I updated the `before_install` hook to ensure that npm v2 is installed unless current travis VM is running node V5 because node V5 comes with npm v3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,6 @@ services:
    - rabbitmq
 
 before_install:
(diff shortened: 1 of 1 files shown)


squaremo:
This is a great improvement! Thank you.

revington:
You're welcome :)

PR #230, opened 2016-02-08
Add example send.js with generators version

rudijs:
examples/send.js done with generators using Co async control flow.

The example works fine for me, though it's fresh cut code that's not yet been used in production.

I think it's handy to have, I understand if you prefer to not add it into the examples/ for whichever reason.

Cheers.


squaremo:
This is a nice example -- thank you. Generators certainly make the code more pleasant don't they!

PR #240, opened 2016-04-18
Better Promise Example in README

woozyking:
The original Promise API example in README has a few exhibition of what's now considered to be [Promise bad practices](https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html), namely:

1. Not utilizing the otherwise elegant feature of "chained-then" pattern.
2. (Also related 1.) Not utilizing `catch` and instead utilizing somewhat inferior `rejectHandler` interface.

In the future, it'll be a great user-experience boost if the `callback_api` can be "merged" with the promise API so that if no callback is given, it automatically returns a promise, like many other libraries do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,28 +83,24 @@ var open = require('amqplib').connect('amqp://localhost');
 
 // Publisher
 open.then(function(conn) {
-  var ok = conn.createChannel();
-  ok = ok.then(function(ch) {
-    ch.assertQueue(q);
-    ch.sendToQueue(q, new Buffer('something to do'));
-  });
-  return ok;
-}).then(null, console.warn);
+  return conn.createChannel();
+}).then(function(ch) {
+  ch.assertQueue(q);
+  ch.sendToQueue(q, new Buffer('something to do'));
+}).catch(console.warn);
 
 // Consumer
 open.then(function(conn) {
-  var ok = conn.createChannel();
-  ok = ok.then(function(ch) {
-    ch.assertQueue(q);
-    ch.consume(q, function(msg) {
-      if (msg !== null) {
-        console.log(msg.content.toString());
-        ch.ack(msg);
-      }
-    });
+  return conn.createChannel();
+}).then(function(ch) {
+  ch.assertQueue(q);
+  ch.consume(q, function(msg) {
+    if (msg !== null) {
+      console.log(msg.content.toString());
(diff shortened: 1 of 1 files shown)


woozyking:
```javascript
// ...
then(function(ch) {
  ch.assertQueue(q);
  ch.sendToQueue(q, new Buffer('something to do'));
}) // ...
```

Note that _side effects_ like above are still considered less desired practices. But I have no domain knowledge about the internals of these channel methods, it'll be good to have this discussed and maybe bettered.

squaremo:
Thank you! Yes, these were written a while ago, and in my own idiosyncratic interpretation of how promises ought to be used.

> In the future, it'll be a great user-experience boost if the callback_api can be "merged" with the promise API so that if no callback is given, it automatically returns a promise

I would rather keep them separate -- I don't think mixing promises and callbacks would result in better code. Maybe that's a lack of imagination on my part though!

woozyking:
It's not about mixing per se. I agree that the usage of promise mixing
callbacks within the scope of, say a function, would be disastrous.

However what I meant is that it would offer great flexibility for usage
within the scope of a module, applying either promise or callback APIs with
one import/require, in different functions (which may have existing code or
other library dependencies that only offer one of the two interfaces).

The implementation would be actually simple in most cases. A fairly well
written example can be found here
https://github.com/mongodb/node-mongodb-native/blob/2.1/lib/mongo_client.js
at where `MongoClient.connect` method is defined.

Of course, that'd be purely an user experience bonus. For now I could
easily import both APIs within the same module if I needed to achieve
aforementioned effect.
Michael Bridgen <notifications@github.com>于2016年4月19日 周二02:18写道：

> Thank you! Yes, these were written a while ago, and in my own
> idiosyncratic interpretation of how promises ought to be used.
>
> In the future, it'll be a great user-experience boost if the callback_api
> can be "merged" with the promise API so that if no callback is given, it
> automatically returns a promise
>
> I would rather keep them separate -- I don't think mixing promises and
> callbacks would result in better code. Maybe that's a lack of imagination
> on my part though!
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/squaremo/amqp.node/pull/240#issuecomment-211753177>
>
-- 
--
Runzhou Li (Leo)

PR #242, opened 2016-05-01
Allow node 6

posgarou:
Test suite passes. I also used it in a simple local project on node v6 and didn't run into any problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/squaremo/amqp.node.git"
   },
   "engines": {
-    "node": ">=0.8 <5 || ^5"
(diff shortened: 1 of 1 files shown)


posgarou:
The Travis failures are due to timeouts unrelated to this change—node current (v6) ran fine.

I didn't explicitly add node v6 to [.travis.yml](https://github.com/squaremo/amqp.node/blob/master/.travis.yml), but Travis runs it because of line 12 in the travis file.  Happy to add it explicitly if you like.

squaremo:
Thanks!

(Re-ran the Travis build to be sure)

PR #262, opened 2016-07-27
Updates the url that links to rabbit_heartbeat.erl

tapickell:
I know this is not pressing but when I was going through this project
trying to learn more about the heartbeat part of RabbitMQ I noticed
that this link was no longer valid as that part has been moved out into
the rabbitmq-commom package.

Todd Pickell @tapickell <tapickell@gmail.com>


squaremo:
Good spot, thanks Todd!

PR #283, opened 2016-09-22
Update callback_model.js

shanksauce:
Probably didn't mean to assign f.id?


squaremo:
Oops! Good catch, thank you.

PR #289, opened 2016-09-30
Browser WebSocket-tunneled AMQP client

shanksauce:
This is a rough, first-pass attempt to get amqp.node working in a browser. It requires the use of a WebSocket server proxying upstream traffic to the AMQP broker. Additionally, it requires the use of browserify in order to "compile" the client code. Someday it'd be nice to remove this dependency.


shanksauce:
https://www.youtube.com/watch?v=BWQACEMW0i4

squaremo:
Interesting -- I was always sure it would be more or less impossible to make this work in a browser! Thank you for what must have been a bunch of effort getting this to work.

Disclaimer: I'm finding it quite difficult to keep up with amqplib issues and PRs and email queries and so on, so please accept my retroactive apologies for sitting on things more or less indefinitely.

Some questions:
 * what's needed at the server end to make this work? I wasn't aware that RabbitMQ supported websockets, so presumably some proxy is needed?
 * ws_connection.js is a near facsimile of connection.js, is the intention that they end up sharing code?
 * will it work ex-browser, e.g., with a suitably prepared websocket library?

shanksauce:
Appreciate your response! Sorry for my own delayed response. No worries. 

- AFAIK, RabbitMQ offers STOMP and MQTT plugins, both exposing a WebSocket server on the broker that can be utilized via a web browser. 

  https://github.com/rabbitmq/rabbitmq-web-mqtt

  In contrast, my intention for this little experiment was to both 1) utilize AMQP 0-9-1 (with RabbitMQ extensions) directly from web browser client code, and 2) to avoid reconfiguring the broker. In production, I prefer to avoid binding RabbitMQ on any WAN interfaces in the first place, so to answer your first question, yes, a tunneling proxy (`server.js` in my PR) is required in this case. 

- Indeed - I think there are a variety of approaches one could take towards encapsulating the functionality. My PR is hacked together, so I didn't want to pollute good code. However if a more serious effort is placed on integration, then there is definitely a lot of boilerplate that could be shared between the browser and NodeJS code, the most notable differences at least from my perspective being related to WebSocket interfaces, reading/writing binary data (FileReader, UInt8Array, ArrayBuffer <-> Buffer). A little leveraging of Strategy Pattern here would go a long way. 

- I don't see why not - WebSockets are only being utilized for full-duplex TCP in the browser. RFC6455 just comes with the territory, and is not being used beyond an encapsulation mechanism per se, so any WebSocket library should suffice. Personally, I can't think of good reasons to use WebSockets outside of a web browser as we have far richer networking capabilities available to us in system environments.

It's been a long time since I looked at this PR and eventually I should dust it off and catch up to the latest changes in the core library. Right now all this really represents is how cool I think it would be to see AMQP used more in the wild, particularly in web applications.

cressie176:
Tidy up

PR #295, opened 2016-11-01
Migrate when -> bluebird

nfantone:
This is a new take on #158. I realize, since that PR has been seating there for a year and a half now, it won't ever be updated.

I won't add much to what was already commented on #158, except that this time around everything seems to be working fine, including unit tests. No API changes, AFAIK. This should be completely transparent to users. It also gets rid of the so-called ["deferred anti-pattern"](http://bluebirdjs.com/docs/anti-patterns.html), widely used in the library.

### DIFF
--- a/channel_api.js
+++ b/channel_api.js
@@ -1,14 +1,14 @@
 var raw_connect = require('./lib/connect').connect;
 var ChannelModel = require('./lib/channel_model').ChannelModel;
-var defer = require('when').defer;
+var Promise = require('bluebird');
 
 function connect(url, connOptions) {
-  var opened = defer();
-  raw_connect(url, connOptions, function(err, conn) {
-    if (err === null) opened.resolve(new ChannelModel(conn));
-    else opened.reject(err);
+  return Promise.fromCallback(function(cb) {
+    return raw_connect(url, connOptions, cb);
+  })
+  .then(function(conn) {
+    return new ChannelModel(conn);
   });
-  return opened.promise;
 };
 
 module.exports.connect = connect;
--- a/examples/tutorials/README.md
+++ b/examples/tutorials/README.md
@@ -9,7 +9,7 @@ of the tutorial programs to the callback-oriented API.
 To run the tutorial code, you need amqplib installed. Assuming you are
 in a clone of the amqplib repository, from the tutorials directory:
 
-    npm install ../..
+    npm install
 
 or to use the latest released version,
 
--- a/examples/tutorials/emit_log.js
+++ b/examples/tutorials/emit_log.js
@@ -1,10 +1,9 @@
 #!/usr/bin/env node
 
 var amqp = require('amqplib');
-var when = require('when');
 
 amqp.connect('amqp://localhost').then(function(conn) {
-  return when(conn.createChannel().then(function(ch) {
+  return conn.createChannel().then(function(ch) {
     var ex = 'logs';
     var ok = ch.assertExchange(ex, 'fanout', {durable: false})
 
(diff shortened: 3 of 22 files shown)


michaelklishin:
You can verify that [rabbitmq-tutorials](https://github.com/rabbitmq/rabbitmq-tutorials/) are still functional with this change just to gather another data point :)

nfantone:
@michaelklishin Good idea. Also, you made me realize that I didn't remove `when` from the examples on _this_ repo. I'll add some more commits in a minute.

nfantone:
@michaelklishin Examples on [rabbitmq-tutorials](https://github.com/squaremo/amqp.node/blob/master/examples/tutorials) seem to be the same as the ones included in [this repo](https://github.com/squaremo/amqp.node/tree/master/examples/tutorials). So, they should work just fine.

nfantone:
@squaremo Tests running on node 4 threw a `TIMEOUT` error while trying to connect to the broker. Every other job in the matrix passed just fine.

screen shot 2016-11-01 at 12 39 07 pm

If you retry the build, it should be good to go.

Also, you should _really_ consider:

* Dropping support for io.js and node 0.8. Both [completely unsupported by now](https://github.com/nodejs/LTS#lts-schedule).
* Run tests on node 6 LST.

nfantone:
@squaremo Also, if you worry about users that may have been using `when` specific, non-standard, Promise methods, you should know that all of its API is defined in `bluebird` one way or another. Conversion is trivial.

Here's a quick "migration guide" from [when](https://github.com/cujojs/when/blob/master/docs/api.md) -> [Bluebird](http://bluebirdjs.com/docs/api-reference.html):

`when.done(handleResult, handleError)` -> `.done`
`when.then(onFulfilled)` -> `.then`
`when.spread(onFulfilledArray)` -> `.spread`
`when.fold(combine, promise2)` -> `.join`
`when.catch(onRejected)` -> `.catch`
`when.finally(cleanup)` -> `.finally`
`when.yield(x)` -> `.return`
`when.else(x)` -> `.catchReturn`
`when.tap(onFulfilledSideEffect)` -> `.tap`
`when.delay(milliseconds)` -> `.delay`
`when.timeout(milliseconds, reason)` -> `.timeout`
`when.inspect()` -> `.isRejected / .isFulfilled / .isPending / .isCancelled`
`when.with(thisArg)` -> `.bind`
`when.progress(onProgress)`  (deprecated API)

squaremo:
I bit the bullet :tada: I think this warrants a 0.5 release (a minor version bump <v1 is like a major version bump >v1 right?).

Many thanks @nfantone, and to @myndzi for the first time around. And to everyone for being patient, or at least quiet :)

nfantone:
@squaremo Yey! `0.5.0` seems fine to me, as changes _are_ backwards compatible with minor tweaks. And those changes should only apply if you have been using `when`specific extensions.

wtgtybhertgeghgtwtg:
It's a little late to say something, but wouldn't it be a better time to just the minimum required node version to 0.12 and use native promises, since support for node 0.10 ended today?

nfantone:
1. @wtgtybhertgeghgtwtg, sir, you've awarded most unpronounceable name in all Github kingdom.
2. `bluebird` >>> native `Promise`. My opinion (and also others, see [this](http://bluebirdjs.com/docs/benchmarks.html) and [this](http://stackoverflow.com/questions/34960886/are-there-still-reasons-to-use-promise-libraries-like-q-or-bluebird-now-that-we))

wtgtybhertgeghgtwtg:
It's a dependency that duplicates native functionality for negligible and [dwindling](http://v8project.blogspot.com/2016/07/v8-release-53.html) performance benefits.

nfantone:
No, it's not.

But you are certainly free to open a new PR with your proposed changes.

wtgtybhertgeghgtwtg:
If such a PR would be accepted, I would be happy to do so.

PR #296, opened 2016-11-02
Add typescript type definitions

nfantone:
@squaremo I don't know if you are really into this - but _hey_, here it is, nonetheless.

This PR would allow library/module writers to consume "official" type definitions for `amqplib` in their TS 2.0 projects (such as [rx-amqplib](https://github.com/SkippyZA/rx-amqplib)) without hassle, by just installing the client from `npm`.

```sh
npm i amqplib
npm i --save-dev @types/bluebird @types/node
```

Additional `"peerDependencies"` are needed because the definitions file (`index.d.ts`) depends on external typings. Regular users won't need to install those and can safely ignore all this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,18 +13,25 @@
   },
   "dependencies": {
     "bitsyntax": "~0.0.4",
+    "bluebird": "^3.4.6",
     "buffer-more-ints": "0.0.2",
-    "readable-stream": "1.x >=1.1.9",
-    "bluebird": "^3.4.6"
+    "readable-stream": "1.x >=1.1.9"
   },
   "devDependencies": {
-    "mocha": "~1",
     "claire": "0.4.1",
-    "uglify-js": "2.4.x",
-    "istanbul": "0.1.x"
+    "istanbul": "0.1.x",
+    "mocha": "~1",
+    "typescript": "^2.0.6",
+    "uglify-js": "2.4.x"
   },
+  "peerDependencies": {
+    "@types/bluebird": "^3.0.35",
+    "@types/node": "^6.0.46"
+  },
+  "types": "./index.d.ts",
   "scripts": {
-    "test": "make test",
+    "test:typings": "tsc index.d.ts --noEmit",
+    "test": "npm run test:typings && make test",
     "prepublish": "make"
   },
   "keywords": [
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,6 @@ services:
 
 before_install:
    - node --version | (grep  -v 'v5' && npm install -g npm@2) || true
+   - npm i @types/bluebird @types/node
 
 sudo: false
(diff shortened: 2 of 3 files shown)


squaremo:
This looks worthwhile. I am a little worried that it means people changing the JS code will need to also change the type definitions, or they will be out of date (i.e., it will require some expertise in TypeScript). Can you comment on that?

package.json
```diff
@@ -13,18 +13,25 @@
   },
   "dependencies": {
     "bitsyntax": "~0.0.4",
+    "bluebird": "^3.4.6",
```
nfantone: Sorry for the shuffling here. Seems like `npm i --save` sorted entries alphabetically.

nfantone:
@squaremo Well... the answer is: it depends. If changes on JS code update the API, TS definitions should be changed as well to reflect that. No way around that, I'm afraid.

> it will require some expertise in TypeScript

Typescript is very easy to grasp. And we are only interested in typings here. Just by looking at the only relevant file, `index.d.ts`, you can already get a feel of what's going on. No bones about it, really.

If this is a concern for you, people can always install those using `typings` via [DefinitelyTyped repos](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/amqplib). Microsoft is trying to move users away from that, though.

Your call.

lautarodragan:
+1
This would add a lot of value to us TS users.

The "ideal" way proposed by the TS team is for projects to offer their own typings. This gives them more visibility and helps to keep them updated, since the typings would be "official".

The alternative is having them in the [DefinitelyTyped repo](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/amqplib) (the url changed) and publishing them to [@types/amqplib](https://www.npmjs.com/package/@types/amqplib), installing them with `npm i -D @types/ampqlib`.

As @nfantone said, the typings would only need to be updated if the API changes.

kibertoad:
@squaremo What is the final decision on that, do we want it in? I can fix the conflict in this PR if it is OK in general.

MrManny:
It has been over a year. Any progress on this?

squaremo:
Having used TypeScript in [another context][jkcfg/jk], I have a better appreciation of why type definitions are useful. Can we check the types are correct in the CI build?

lautarodragan:
@squaremo you can't check that the types match the implementation if the source code isn't written in TypeScript itself, but there are some basic tests that can be done. 

Basically declaring stuff and seeing it pass the type checker.

Here's ReactJS type definitions tests, for example: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v15/test/index.ts

Recodify:
Any chance we can get this merged?

zernie:
@squaremo, @nfantone sorry, have this been abandoned? Do you need some help?

daniel-pedersen:
There is [@types/amqplib](https://www.npmjs.com/package/@types/amqplib). This has worked for our purposes. But official support would be appreciated. I can also help out if needed.

cressie176:
Maintaining the types in this library would add too much of a burden I'm afraid. It's best they stay at @types/amqplib

Issue #299, opened 2016-11-02
expose IllegalOperationError


ben-page:
I've developed a small library that uses amqplib to make it easy to perform RPC over a message bus. I need to be able to distinguish between errors caused by programmer error while using the library and errors that are caused bugs in the library itself. Testing if an error is an instance of IllegalOperationError is very helpful in this.

```javascript
const IllegalOperationError = require('amqplib/lib/error').IllegalOperationError;
```
Currently, I am accessing IllegalOperationError this way. But circumventing the public API in this way is worrisome, because it could break just by reorganizing or renaming files.

It seems like good practice to expose custom error types in the public API. And it seems to be common for Nodejs libraries to do so.

Uzlopak:
@squaremo 
This issue could be closed as the PR is already merged.

PR #300, opened 2016-11-02
exposes IllegalOperationError to public API

ben-page:
#299 

```javascript
const amqp = require('amqplib');
const IllegalOperationError = amqp.IllegalOperationError;
```
or
```javascript
const amqp = require('amqplib/callback_api');
const IllegalOperationError = amqp.IllegalOperationError;
```


squaremo:
Yep, seems reasonable. Thank you!

PR #304, opened 2016-11-10
Specify heartbeat with object configuration

### LINKED ISSUE #229: No existing functionality to pass in query params in url object
amqp support query string params to allow further options on the connection. They are only available with `amqplib` if your param type is a url connection string; if you pass in an object, there is no support for the query string as an object.
### Missing functionality

`frameMax`
`channelMax`
`heartbeat`
`locale`

ikhemissi:
Provide extra settings inside the object configuration that were only available from url configs
like heartbeat interval.

Closes #229

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -118,7 +118,14 @@ function connect(url, socketOptions, openCallback) {
       pass = url.password;
     }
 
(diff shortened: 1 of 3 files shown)


squaremo:
Thanks for spotting and filling this lack @ikhemissi :)

mikael-entefy:
When will this be published?

PR #330, opened 2017-02-23
Split the username and password correctly

dietervanhoof:
Split the username and password correctly so that the password can contain a ':' character.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -80,7 +80,7 @@ function openFrames(vhost, query, credentials, extraClientProperties) {
 function credentialsFromUrl(parts) {
   var user = 'guest', passwd = 'guest';
   if (parts.auth) {
-    var auth = parts.auth.split(':');
+    var auth = parts.auth.split(/:(.+)/);
     user = auth[0];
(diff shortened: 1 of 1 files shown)


squaremo:
Oops, that's a good spot. Instead of the split on regexp, which I must admit I didn't grok on first scan, would it be better to use `indexOf`, wdyt?

dietervanhoof:
Apparently indexOf would be a tiny bit faster when using smaller strings compared to a regex. I wouldn't mind it using indexOf but I feel like this code is a lot cleaner than something like

```
user = parts.auth.substring(0, parts.auth.indexOf(':'));
pass = parts.auth.substring(parts.auth.indexOf(':') + 1, parts.auth.length)
```

It's all up to you I suppose.

squaremo:
I did end up rewriting it around `indexOf(':')`, mainly because there's a special case (no colon) that I wanted to be explicit.

Thanks for this PR, it fixed the problem but it _also_ made me rethink some of the surrounding bits, revisit the specs, and add some tests!

PR #345, opened 2017-04-06
Respect heartbeat=0 configuration - Fixes #344

### LINKED ISSUE #344: heartbeat=0 connection option is ignored
Using `"amqplib": "^0.5.1"`

Repro by setting `heartbeat=0` and log internal `connection.heartbeat` value:
```
let open = amqp.connect("amqp://guest:guest@127.0.0.1:5672?heartbeat=0");
open.then((conn) => {
      return conn.createChannel();
    }).then((ch) => {
      console.log("amqp hb", ch.connection.heartbeat);
      // => amqp hb 60
});
```

Workaround by explicitly clearing the heartbeat after connecting:
```
...connect, then...
ch.connection.heartbeater.clear();
```

UnbounDev:
Keeping w/ the API docs:

> `heartbeat`: the period of the connection heartbeat, in seconds. Defaults to 0, meaning no heartbeat. OMG no heartbeat!

Allow `heartbeat=0` to be set. Fixes #344.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -225,7 +225,7 @@ C.open = function(allFields, openCallback0) {
       tunedOptions.channelMax =
         negotiate(fields.channelMax, allFields.channelMax);
       tunedOptions.heartbeat =
-        negotiate(fields.heartbeat, allFields.heartbeat);
+        (allFields.heartbeat === 0 ? 0 : negotiate(fields.heartbeat, allFields.heartbeat));
       send(defs.ConnectionTuneOk);
       send(defs.ConnectionOpen);
       expect(defs.ConnectionOpenOk, onOpenOk);


michaelklishin:
@UnbounDev I'm not sure if it's a good idea to disable heartbeats but technically since the protocol allows for it, so should client libraries. @squaremo WDYT?

squaremo:
> @squaremo WDYT?

From https://github.com/squaremo/amqp.node/pull/83#issuecomment-50988301:

> [...] Clients are meant to have the following tuning logic:
>
>  -   when server sends 0, reply with the configured client value
>  -   otherwise, when the configured client value is 0, reply with sent server value
>  -   otherwise, reply with min of client and server value

In this PR, the server value is ignored if the client value is `0`, which breaks the second rule above. This will either result in RabbitMQ cutting off the client because it failed to follow the rules, or the client and the server having different ideas of what the heartbeat is (which, I would guess, will lead to RabbitMQ cutting off the client because it misses a heartbeat).

The whole tuning thing is rather mind-bending, as the full thread of that other PR will attest. I should really fix the documentation, which strongly suggests that it's possible to just switch heartbeats off. (That's only true if the server has `0` as the default -- RabbitMQ doesn't).

squaremo:
#379 fixed the documentation so it no longer implies you can (unconditionally) switch heartbeats off by supplying `0`. (see comments above about why this PR can't be merged -- sorry!)

PR #350, opened 2017-05-11
Adapt to new Buffer interface

furstenheim:
Old syntax for Buffer is being deprecated. I've added the dependency to safe buffer so that the new methods can be used. All tests run nicely in node 4 and node 7


squaremo:
Thank you for doing all this work @furstenheim!

michaelklishin:
@furstenheim thank you!

Issue #358, opened 2017-06-04
Allow explicit number type declaration


pcan:
I'm currently trying to send messages from amqp.node to a Java Spring Application, that expects an header as `long`, see this listener for example:

```Java
@RabbitListener(queues = "testQueue")
public void handleRequest(@Payload RequestBean bean, @Header("trace-id") long traceId) {
    // some logic here...
}
```

I'm getting a type mismatch exception when the `handleRequest` method gets called, since the expected type for `trace-id` is `long` (64-bit integer), but it arrives as `byte` (8-bit integer). I found that this behaviour is implemented [here](https://github.com/squaremo/amqp.node/blob/master/lib/codec.js#L163-L178), and at the moment there is no way to force the number type from outside. 

I think the codec may handle these special cases like it does for `timestamp` (see [here](https://github.com/squaremo/amqp.node/blob/master/lib/codec.js#L107-L111) and [here](https://github.com/squaremo/amqp.node/blob/master/lib/codec.js#L183-L185)), by adding `'int8'`, `'int16'`, `'int32'` and `'int64'` to that switch-case block, and let the user choose the right integer type if needed. The existing heuristic may not be changed, to avoid a breaking change.
So, I may write something like this:

```javascript
var options = {
	headers: {
		'trace-id': {'!': 'int64', value: 12}
	},
	contentType: 'application/json',
	contentEncoding: 'UTF-8'
};

channel.publish('myExchange', 'routing.key', new Buffer(JSON.stringify(message)), options);
```

squaremo:
I like this idea. You're right that it need not interfere with the existing encodings. I'm not sure there are _signed_ 32 bit integers in AMQP 0.9.1 (or RabbitMQ's codec, anyway). But hence or otherwise, this would be fairly easy to add. PR? :)

michaelklishin:
This is not a bad idea. Numerical type interop between languages is a pain and sometimes more control is the only way to achieve it.

On a related note, this is one of the reasons to avoid using message headers where possible and
stick to serialised message bodies.

pcan:
Thanks. I will submit a PR asap.

PR #359, opened 2017-06-14
Fix short integer encoding

Arkq:
Messages encoded with this library are not conformant with AMQP specification - such an incorrect message might crash client (written in other language, in my case py-amqp-1.4.9) which is conformant with the AMQP 0-9-1 specification.

In short: `s` is for short string, not for short integer. For signed short integer is `U`.

Reference:
https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf (Technical Specifications, page 31)

PS.
This commit also changes signedness of the long integer. From now, every integer is encoded as a signed one.

### DIFF
--- a/lib/codec.js
+++ b/lib/codec.js
@@ -165,7 +165,7 @@ function encodeFieldValue(buffer, value, offset) {
                 buffer.writeInt8(val, offset); offset++;
             }
             else if (val >= -0x8000 && val < 0x8000) { //  short
-                tag('s');
+                tag('U');
                 buffer.writeInt16BE(val, offset); offset += 2;
             }
             else if (val >= -0x80000000 && val < 0x80000000) { // int
@@ -173,7 +173,7 @@ function encodeFieldValue(buffer, value, offset) {
                 buffer.writeInt32BE(val, offset); offset += 4;
             }
             else { // long
-                tag('l');
+                tag('L');
                 ints.writeInt64BE(buffer, val, offset); offset += 8;
             }
         }
@@ -215,6 +215,7 @@ function decodeFields(slice) {
         var tag = String.fromCharCode(slice[offset]); offset++;
         switch (tag) {
         case 'b':
+        case 'B':
             val = slice.readInt8(offset); offset++;
             break;
         case 'S':
@@ -222,6 +223,12 @@ function decodeFields(slice) {
             val = slice.toString('utf8', offset, offset + len);
             offset += len;
             break;
+        case 's':
+            len = slice.readUInt8(offset); offset++;
+            val = slice.toString('utf8', offset, offset + len);
+            offset += len;
+            break;
+        case 'i':
         case 'I':
             val = slice.readInt32BE(offset); offset += 4;
             break;
@@ -251,9 +258,11 @@ function decodeFields(slice) {
             val = slice.readFloatBE(offset); offset += 4;
             break;
         case 'l':
+        case 'L':
             val = ints.readInt64BE(slice, offset); offset += 8;
             break;
-        case 's':
+        case 'u':
+        case 'U':
             val = slice.readInt16BE(offset); offset += 2;
             break;
         case 't':


michaelklishin:
@Arkq are you aware of the [errata document](http://www.rabbitmq.com/amqp-0-9-1-errata.html)? The 0-9-1 spec is known to have inconsistencies and contradictions when it comes to table encoding. At this point client library maintainers are better off by following what the Java (and .NET, and Ruby, and many other) clients use than blindly following the spec.

The prefix for short signed integers [indeed seems to be an `s`](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/ValueWriter.java#L182) there.

michaelklishin:
I strongly oppose this change. What needs to change is `py-amqp` (or you should move to Pika if you can), not most other clients. There is no right or wrong way of encoding table values but the errata document was produced because the spec has contradictions that were never clarified (the work then switched to AMQP 0.10, then 1.0).

Here's a list of popular clients that encode strings using `S` and `s` (or `I`, depending on what numerical types are supported or common in a particular language) for short signed integers:

 * [Java client](https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/ValueWriter.java#L128)
 * [.NET client](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/5d0377b4b0cf6e640ba2477f54ee76787813ccdf/projects/client/RabbitMQ.Client/src/client/impl/WireFormatting.cs#L274)
 * [Bunny](https://github.com/ruby-amqp/amq-protocol/blob/master/lib/amq/protocol/type_constants.rb)
 * [Pika](https://github.com/pika/pika/blob/51fd186f1fabe0da8582609d266df1e28e57e3f4/pika/data.py#L91)
 * [php-amqplib](https://github.com/php-amqplib/php-amqplib/blob/a59bb7fbda76127f812a72b15cf44bd54421d506/PhpAmqpLib/Wire/AMQPWriter.php)

By introducing this change you will break this client's interoperability with those client and gain interoperability with `py-amqp`. The loss would be much greater than the gain if you ask me.

@squaremo would you agree with me?

Arkq:
@michaelklishin thanks for the errata link. I was not aware of this document. During my battle with this strange bug (encoding incompatibilities), I've stumbled upon 0.9.1 spec and I've thought that it is the "master". If you say, that Pika is a way to go, I will see if it is possible in my case.

Many thanks for such a quick response!

I'm closing this PR as invalid.

michaelklishin:
@Arkq thank you. I completely understand that this is very confusing. The AMQP 0-9-1 spec had a rough childhood.

squaremo:
@michaelklishin Yes I agree; the codec here is carefully crafted to be compatible with RabbitMQ and do what (most) other client libraries do.

Unfortunately, the reason that one has to take a lot of care also means that it's not at all obvious why it's the way it is!

Arkq:
After some investigation, I have learned, that [pika is NOT a way to go](https://github.com/pika/pika/blob/b907f91415169b7f590174ab5d228e75a1b273e6/pika/data.py#L255) (pika-0.10.0) if one wants interoperability with clients conformant with AMQP 0.9.1 + errata. Unfortunately, that was after, I've ported everything to pika... :)

So, for anyone who has stumbled upon similar issue, the solution is:
* [haigha](https://github.com/agoragames/haigha/blob/master/haigha/reader.py#L382)
or
* [amqpstorm](https://github.com/eandersson/amqpstorm) (due to usage of [pamqp](https://github.com/gmr/pamqp/blob/067679dab0c1ee2d02dd6dbdcd5c5f29ea1e9e9c/pamqp/decode.py#L292))

PS.
I don't know if anyone who maintains rabbitmq developer tools website (https://www.rabbitmq.com/devtools.html) reads these comments, but if so, I would recommend to investigate, which client is conformant which which specification. This might help people like me in the future.

PS2.
Currently I'm using: [amqp.node](https://github.com/squaremo/amqp.node) for Node, [haigha](https://github.com/agoragames/haigha) for Python, [amqp](https://github.com/streadway/amqp) for Go, and [spring-amqp](https://github.com/spring-projects/spring-amqp) for Java. Every library seems to cooperate with each other as expected.

michaelklishin:
We intend to work on Pika instead of moving the tutorials for other clients. Pika is certainly fine as far as interoperability goes in most places. Attribute tables was or still is an edge case in most clients.

Arkq:
@michaelklishin that's fine. Tutorial is only an example (imho). It could have been written in a pseudocode as well, it would serve its purpose. However, I'm not complaining about the tutorial, but about the list of RabbitMQ libraries which is kinda misleading (for edge cases).

_\<cool-story>_
For a past few years I (actually the company I'm working for) have been using kombu with py-amqp. There was no problems, because we've been using mostly Python for rabbitmq-related work. Even recently, when the variety of used languages has expanded (including amqp.node), we have encountered no problems at all. Because our RabbitMQ usage was kind of "standard". Send and receive message without any customizations. However, few days ago I've added custom headers while sending messages and... kabooom. Amqp.node has encoded integer value 1000 as a short integer "s", which has destroyed our ecosystem :D
So, for most cases these libraries are fine, but someday....
_\</cool-story>_

Maybe the solution would be to change the amqp.node after all. E.g.: Encode integers in a compatible way with all clients - as an integer. Decoding will stay as it is now. This would be similar to Pika - which for encoding uses an intersection of types from 0.9.1 and 0.9.1-errata. In such a way there is 100% guaranteed compatibility with all clients.

squaremo:
> However, few days ago I've added custom headers while sending messages and... kabooom. Amqp.node has encoded integer value 1000 as a short integer "s", which has destroyed our ecosystem :D

Since JavaScript only has the one numerical type, double, I made a guess about what was the most useful way to encode numerical values: by prioritising precision, then size of encoding, more or less. But it's impossible to get this right for everyone. (Note for protocol designers: don't include a "generic" encoding for values.)

There's a suggestion that may turn into a PR: https://github.com/squaremo/amqp.node/issues/358. This would let you specify that a value is specifically a long, or short, or whatever. Possibly I ought to have just required people to do this all along! (But then people using other dynamic languages would have found it less convenient, ho hum).

Arkq:
> Possibly I ought to have just required people to do this all along!

For JS it would be a little bit strange. Like you've said, there is only one numerical type. However, it is possible to determine whether numerical value is an integer or not. So this approach of encoding floats and integers in a separate way is OK (IMHO).

I've seen issue #358 before. It would be possible to mitigate my problem by using integer (int32). And to be fair, we've got exactly the same problem as described in the issue #358. In fact this problem appears in Go as well.

However, right now I don't care much about the encoding. Changing Python library has resolved my issues. And the problem with dynamically selected integer type is quite easy to overcome anyway.

I will not bump this thread any more. Once again many tanks to @michaelklishin, who has pointed me in a right direction - amqp spec incompatibilities.

michaelklishin:
@Arkq that's borderline overkill and an overly smart solution. Dynamically typed languages will always struggle with the plethora of numerical types available in/to their statically typed counterparts. JavaScript is particularly limited in this area.

Unless you have to use headers exchanges (which is an exception, not the norm), I'd recommend sticking your values into message payload, and encoding the entire thing using protocol buffers.

I think the idea in #358 is quite good, given the circumstances.

Issue #366, opened 2017-07-26
Sending message with large header throws exception


twatson83:
I need to be able to pass a large encrypted string as a header when sending messages.  Below is the code used to send the message,

```
this.channel.sendToQueue(ep, message, { headers: messageHeaders, messageId: messageId });
```

And here is the exception that is thrown.

```
RangeError: Index out of range
    at checkInt (buffer.js:1045:11)
    at Buffer.writeUInt8 (buffer.js:1091:5)
    at encodeTable (codec.js:83:18)
    at encodeBasicProperties (defs.js:2773:11)
    at module.exports.encodeProperties (defs.js:3544:12)
    at Connection.C.sendMessage (connection.js:542:16)
    at Channel.C.sendMessage (channel.js:85:26)
    at Channel.publish (callback_model.js:171:17)
    at Channel.sendToQueue (callback_model.js:175:15)
```

I believe the issue is here 

https://github.com/squaremo/amqp.node/blob/master/bin/generate-defs.js#L125

Increasing the SCRATCH buffer size fixes the issue, but maybe this value shouldn't be hard coded?

michaelklishin:
@twatson83 bumping the value to e.g. 16 kiB should be an OK stop-gap measure. Please submit a PR? Also, what's the size of your string value?

twatson83:
The string is 10836 characters long.  Ok I will submit a PR.

twatson83:
This is now an issue again.  

We store a JWT in the header which is now bigger the 16KiB for some users.

Can we increase the size to 32KiB?

PR #367, opened 2017-07-26
Increased SCRATCH buffer to 16 KiB

twatson83:


michaelklishin:
This is a follow-up to #366. Thank you, @twatson83.

@squaremo I think this change is reasonable and should cover a lot of scenarios where large headers are used. I'd prefer to use this buffer size without making it configurable for some time, until we have evidence that enough users need it. WDYT?

squaremo:
Hi Tim!

No objection to this -- it's a one-off buffer, so the size doesn't matter too much within bounds ("4K should be enough for anyone!").

PR #378, opened 2017-09-25
Added flags closing and closed to connection and channel

KristjanTammekivi:
Since order of messages is not guaranteed when using multiple channels I have a use case where I want to keep a single channel open as long as possible and only create a new one when the old one is closed. I could use hack ways to detect that but I'd rather have explicit flags for that.

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -35,6 +35,8 @@ function Channel(connection) {
   }));
   // message frame state machine
   this.handleMessage = acceptDeliveryOrReturn;
+  this.closed = false;
+  this.closing = false;
 }
 inherits(Channel, EventEmitter);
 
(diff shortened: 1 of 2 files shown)


squaremo:
Thanks @KristjanTammekivi :) There are events for channels and connections being closed (admittedly, there are none for closing). Do you have example code which shows these flags in use?

I'm not against having these flags, just wary of having two ways of doing things.

cressie176:
Hi @squaremo. I'd find this pull request useful. It would save me [decorating channels with my own attributes](https://github.com/guidesmiths/rascal/blob/master/lib/amqp/Vhost.js#L120-L125) in rascal.

NaughtyMuppet:
The reason I don't want to use the events is that they're asynchronous and I'm scared of race conditions.

KristjanTammekivi:
I don't want to use events because they're asynchronous and I'm scared of race conditions.

KristjanTammekivi:
TIL EventEmitter is synchronous, not async as I originally thought. Therefore this is not needed, closing.

PR #381, opened 2017-10-05
Add method for ConfirmChanne#publish and ConfirmChanne#sendToQueue

jwgoh:
The document mentions that `ConfirmChannel#publish` and `ConfirmChannel#sendToQueue` does not return promises in the promise API under [flow control](http://www.squaremobius.net/amqp.node/channel_api.html#flowcontrol) but not under the section titled **ConfirmChannel**


squaremo:
Thank you! :star:

Issue #388, opened 2017-11-10
Connect reports a timeout, but may still return the connection


cressie176:
#### Steps to reproduce

1. start a docker container with rabbitmq server
1. pause the container
1. connect using amqplib
1. wait for a timeout, log the error, but don't crash
1. unpause the container

Initially the timeout will be reported, then after the container is unpaused the connection will be returned

#### Sample Code
```
var amqplib = require('amqplib/callback_api')

amqplib.connect('amqp://localhost', { timeout: 5000 }, function(err, connection) {
  if (err) return console.log(err.message)
  console.log('ok')
})

setTimeout(function() {}, 10000);
```

#### Sample Output
```
node demo.js 
connect ETIMEDOUT
ok
```

This is likely to cause connection leaks for any application or library which attempts to reconnect on error. There are other issues where multiple errors can be reported, causing the callback to be executed multiple times. Would be really good to have these fixed in amqplib rather than workaround them.

cressie176:
Appears fixed in 0.5.2. Thanks.

PR #411, opened 2018-03-23
Connection recovery.

hairyhum:
A way to provide connection recovery in case of network issues.
Currently it recovers connections, channels and consumers.
The PR is work in progress and misses some critical features (like channel configuration recovery).
There are no tests yet.

@squaremo what do you think of this approach, does it make sense to you? I'm not an expert in Node.js and can miss some things. 

I've looked if I can modify the tests to test recovery, but I cannot understand what the "unexpected socket close" is checking. I cannot find any assertions there and if I force-enable connection recovery - the test does not fail. Can you help me understand what the test is asserting?

### DIFF
--- a/lib/callback_model.js
+++ b/lib/callback_model.js
@@ -488,6 +488,7 @@ Channel.prototype.get = function(queue, options, cb0) {
       else if (f.id === defs.BasicGetOk) {
         self.handleMessage = acceptMessage(function(m) {
           m.fields = f.fields;
+          m.incarnation = self.incarnation;
           cb(null, m);
         });
       }
--- a/lib/callback_model.js
+++ b/lib/callback_model.js
@@ -62,7 +62,7 @@ function callbackWrapper(ch, cb) {
   } : function() {};
 }
 
-Channel.prototype.recoverChannel = function(cb){
+Channel.prototype.recoverChannel = function(recoverTopology, cb){
   this.revalidate();
   var self = this;
   var recoverPrefetchAndConsumers = function(error, _ok) {
@@ -73,6 +73,7 @@ Channel.prototype.recoverChannel = function(cb){
         if(error !== null) {
           cb(error, res);
         } else {
+          // TODO recover topology
           self.recoverConsumers(cb);
         }
       });
--- a/lib/callback_model.js
+++ b/lib/callback_model.js
@@ -186,7 +186,7 @@ Channel.prototype.recoverQueues = function(callback) {
   function afterRecovery(self, queue_renames) {
     queue_renames.forEach(function(queue_rename){
       if(queue_rename.old_name !== queue_rename.new_name){
-        self.rename_queue(queue_rename.old_name, queue_rename.new_name);
+        self.renameQueue(queue_rename.old_name, queue_rename.new_name);
       }
     });
     callback(null, self);
@@ -309,7 +309,7 @@ Channel.prototype.assertQueue = function(queue, options, cb) {
                         qname = qok.queue;
                         anonymous = true;
                       }
-                      self.record_queue(qname, anonymous, options);
+                      self.recordQueue(qname, anonymous, options);
                     }
                     cb(error, qok);
                   });
@@ -328,7 +328,7 @@ Channel.prototype.deleteQueue = function(queue, options, cb) {
                   defs.QueueDeleteOk,
                   function(error, delok){
                     if(error === null) {
-                      self.remove_queue(queue);
+                      self.removeQueue(queue);
                     }
                     cb(error, delok);
                   });
@@ -348,7 +348,7 @@ Channel.prototype.bindQueue =
                     defs.QueueBindOk,
                     function(error, bindok) {
                       if(error === null) {
-                        self.record_binding("queue", queue, source, pattern, argt);
+                        self.recordBinding("queue", queue, source, pattern, argt);
                       }
                       cb(error, bindok);
                     });
@@ -362,7 +362,7 @@ Channel.prototype.unbindQueue =
                     defs.QueueUnbindOk,
                     function(error, unbindok) {
                       if(error === null) {
-                        self.remove_binding("queue", queue, source, pattern, argt);
+                        self.removeBinding("queue", queue, source, pattern, argt);
                       }
                       cb(error, unbindok);
                     });
@@ -376,7 +376,7 @@ Channel.prototype.assertExchange = function(ex, type, options, cb0) {
             defs.ExchangeDeclareOk,
             function(e, _) {
               if(e === null){
-                self.record_exchange(ex, type, options);
+                self.recordExchange(ex, type, options);
               }
               cb(e, {exchange: ex});
             });
@@ -396,7 +396,7 @@ Channel.prototype.deleteExchange = function(exchange, options, cb) {
                   defs.ExchangeDeleteOk,
                   function(error, deleteok) {
                     if(error === null) {
-                      self.remove_exchange(exchange);
+                      self.removeExchange(exchange);
                     }
                     cb(error, deleteok);
                   });
@@ -410,7 +410,7 @@ Channel.prototype.bindExchange =
                     defs.ExchangeBindOk,
                     function(error, bindok) {
                       if(error === null) {
-                        self.record_binding("exchange", dest, source, pattern, argt);
+                        self.recordBinding("exchange", dest, source, pattern, argt);
                       }
                       cb(error, bindok);
                     });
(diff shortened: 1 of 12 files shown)


michaelklishin:
@squaremo FYI, some RabbitMQ team members are working on adding connection recovery to this client.

michaelklishin:
@squaremo sorry, any comments on this? Team RabbitMQ would love to introduce connection recovery to this client. Any feedback on this PR and the idea in general?

lib/callback_model.js
```diff
@@ -199,13 +243,33 @@ Channel.prototype.cancel = function(consumerTag, cb0) {
     function(err, ok) {
       if (err === null) {
         self.unregisterConsumer(consumerTag);
+        //TODO: If autorecovery is enabled
+        this.remove_consumer(consumerTag);
```
cressie176: this or self?

lib/callback_model.js
```diff
@@ -199,13 +243,33 @@ Channel.prototype.cancel = function(consumerTag, cb0) {
...
         cb(null, ok.fields);
       }
       else cb(err);
     });
   return this;
 };
 
+Channel.prototype.record_consumer = function(tag, queue, callback, options){
```
cressie176: recordConsumer instead of record_consumer?

lib/callback_model.js
```diff
@@ -199,13 +243,33 @@ Channel.prototype.cancel = function(consumerTag, cb0) {
...
+  this.ensureConsumersBooked();
+  this.book.consumers[tag] = {queue: queue, callback: callback, options: options};
+};
+
+Channel.prototype.remove_consumer = function(tag){
```
cressie176: removeConsumer instead of remove_consumer?

lib/callback_model.js
```diff
@@ -199,13 +243,33 @@ Channel.prototype.cancel = function(consumerTag, cb0) {
...
+  delete this.book.consumers[tag];
+};
+
+Channel.prototype.ensureConsumersBooked = function(){
+  if(this.book == undefined) {
```
cressie176: prefer === 
See https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

lib/callback_model.js
```diff
@@ -199,13 +243,33 @@ Channel.prototype.cancel = function(consumerTag, cb0) {
...
+    this.book = {};
+  }
+  if(this.book.consumers == undefined) {
```
cressie176: prefer ===

cressie176:
When I implemented connection recovery in (https://githubt.com/guidesmiths/rascal)[rascal] I got bitten by the following...

1. When a connection attempted timed out, it might still subsequently succeed. amqplib executed the connect callback twice, the first time with the error, the second time with the connection. This resulted in a connection leak. I believe this was fixed, but it's an interesting case to test for.

1. When the node closed gracefully a close event is emitted. I didn't listen for this initially.

1. Without a timeout it's possible to hammer a node that's restarting if there are multiple consumers. Maybe set a default one?

1. Even with a shared timeout it's possible to initiate thousands of connections at the same time. To avoid this I added a random value to the timeout and added supports exponential backoff.

1. In HA mode users may want to reconnect to a different node. Rascal supports multihost connection strings, randomises the order at startup, then cycles through the hosts on reconnection attempts. May not be appropriate to implement as part of amqplib as similar behaviour can be achieved through DNS.

hairyhum:
@cressie176 thank you for the feedback.

1. I'm not sure how can this happen. Can you explain in more details? In this PR the connection socket is closed before reconnection attempt. I'm not sure if it helps to avoid the problem you are talking about. How can I reproduce?

2. Server-close recovery is disabled by default and can be enabled with a separate config `recoverOnServerClose`. It's not that useful in production, but helps in the tests.

3. Yes. Defaults are a subject to change. There is the `recoverAfter` config to do that.

4. How can I reproduce this? Is that when you have a lot of clients reconnecting at the same time?

5. Connecting to multiple nodes is definitely something we will consider adding. I was trying to make this thing work with a single server first.

cressie176:
@hairyhum thanks for responding. Automatic reconnection & resubscription will be very helpful.

> 1

The issue and steps to reproduce was discussed [here](https://github.com/squaremo/amqp.node/issues/388). Should be fixed but worth testing for.

> 4

Yes - when running a lot of clients. For example we're running around 300 node.js apps with 6 instances each. Not all rely on RabbitMQ, but some connect to multiple VHosts so say 1800 clients for the sake of argument. RabbitMQ is configured in HA mode with 3 nodes. If one of our RabbitMQ nodes has a temporary network issue roughly 600 clients will attempt to reconnect. If they all do this at the same time, whether immediately or after a set timeout there will be a sudden spike in network connections, which we'd rather avoid.

michaelklishin:
@squaremo sorry bot continue bugging you but any chance you can review this and/or grant team RabbitMQ commit/owner bit for this project?

hairyhum:
At this pint most of the recovery features are there and tested. Unfortunately I can't find a way to make management available in Travis CI to make tests pass.

There is a number of additional things, which can be added:

1. Randomized delays. (recoverAfter with random part to avoid reconnect storm)
1. Multiple URLs to connect to in random or round-robin manner.
1. Cross-channel bookkeeping (a queue should not be recovered by a channel if it was deleted from another channel on the same connection)

This things can be added separately. At this point we need to get some feedback from the library users.

@squaremo can you please take a look and merge/comment?

squaremo:
> sorry bot continue bugging you but any chance you can review this 

It doesn't look finished. Since this comment https://github.com/squaremo/amqp.node/pull/411#issuecomment-380873599 suggesting that it was ready to merge, there have been five more commits. And the CI doesn't pass.

I understand that connection recovery is a big ticket item for many people, so it's cool that you are contributing code to make it happen. But my time to spend on amqplib is pretty scarce, so I respectfully ask that you ration your review requests.

@hairyhum Since you've clearly put a lot of energy into this PR to date it seems fair that you get some kind of considered response, finished or no :-)

On the design:

It is my fond hope that connection recovery could be made to compose with the existing code. By that I mean that's it's something you overlay onto the existing API, and you don't have to use it if you don't want to. I'm not suggesting that the existing code and API mustn't be altered; just that it ought to be altered as little as possible, with the minimum hooks and bookkeeping necessary to support the outer, reconnection layer.

Perhaps instead of having a nested if-then-else for the different kinds of recovery, you could implement the whole as a composition of smaller abstractions that each do a kind of recovery. (Maybe you've already considered this approach, and discarded it; I am hand-waving. It would be good to see the rationale for how it's implemented explained in the commit messages, in any case.)

On the commits:

Before I merged this I'd ask for a `git rebase -i` to elide some backtracking, and otherwise rewrite the history into fewer, more cohesive commits. For example, 24aef17 is a fixup of c78569f; the first commit is marked 'WIP'.

michaelklishin:
There are two approaches to connection recovery common in RabbitMQ clients:

 * Provide recovery as a client feature, first as an opt-in, then enabled by default (usually after a few years). Ruby, Java, .NET, Objective-C, Clojure clients do this.
  * Intentionally avoid supporting recovery as a feature. Make it easy to recover manually, provide plenty of examples. Go and now Pika do this.

Neither breaks existing code.

Would a "composable" solution fall into the second category? I'm not sure it's realistic to implement the first option without some of the internals changes from this PR.

hairyhum:
@squaremo thank you for responding.

The commits after the last message were adding multiple hosts support and randomised delays. Also fixing some bugs discovered. It is complete in the sense that I don't plan to add anything more, it is working in the sense that it passes tests if you run them manually and can be tested manually. 

CI fails because I was not able to find a way to communicate with HTTP API which is required in the tests. If there is a way to do that, that would be nice. Otherwise it can be done in a makefile, like in [bunny](https://github.com/ruby-amqp/bunny/blob/master/bin/ci/before_build)

Speaking of composition, do you mean some sort of a wrapper class for recoverable connection (like in Java client)? I can refactor the code to use that approach. There will still be a lot of if-then-elses, because there are configurations to be checked.
I can rebase things as soon as you are happy with the changes.

hairyhum:
Tests show that there are multiple timing related errors in handling TCP errors on different connection states.
The complexity of this implementation makes me think this should be scrapped.
I will work on a simpler solution - a failure callback, which will be called when connection errors.

Taha-Firoz:
Any update on this, I've been scouring around the web looking for what amqplib's reconnect policy and I can't find anything. I have a couple of services consuming on RabbitMQ and want to know what will happen in case Rabbit fails and recovers. Everything's going to be running on Docker with a restart policy, my concern isn't really if my messages can be recovered or not but actually if my consumers will be able to recover and reconnect when RabbitMQ's back up.

PR #413, opened 2018-04-05
 Return a promise-with-ok from ConfirmChannel.publish

canuc:
+ `ConfirmChannel.publish` does not return a promise when it should actually. 
+ Fixed some test

### DIFF
--- a/examples/confirmChannel.js
+++ b/examples/confirmChannel.js
@@ -1,0 +1,19 @@
+var amqp = require('../');
+
+var sent = 0;
+var acked = 0;
+
+amqp.connect().then(function(c) {
+  c.createConfirmChannel().then(function(ch) {
+    var p;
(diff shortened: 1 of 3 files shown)


squaremo:
Thank you for bringing this back! I did not complete it because I had second thoughts about returning something other than a flow-control return value. But I can see why it's useful.

As it stands, this could break code by always returning a truthy value, and (more breakingly) ignoring the callback argument. Perhaps if we made this opt-in by supplying an argument to ConfirmChannel, like:

    conn.createConfirmChannel({returnPromises: true}).then(...)

wdyt?

cressie176:
Are there any other parts of the code base with a similar issue that cannot take a map as either their first or last argument? If so the API would be inconsistent.

Another downside is having to add the `{ returnPromises:true }` on every call.

As an alternative what about a module level setting `preferPromises` or something you specify when creating the connection?

PR #432, opened 2018-06-19
Connection recovery helper.

hairyhum:
The helper function to simplify connection recovery via event emitter.

This function uses a callback, which can be called multiple times, hence cannot be promisified.

TODO:

- [x] Multiple hosts support
- [x] Documentation PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,6 @@ services:
 before_install:
    - node --version | (grep  -v 'v5' && npm install -g npm@2) || true
 
-sudo: false
+before_script: "./bin/ci/before_build"
+
+sudo: required
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ lib/defs.js: $(UGLIFY) bin/generate-defs.js bin/amqp-rabbitmq-0.9.1.json
 		-b 'indent-level=2' 2>&1 | (grep -v 'WARN' || true)
 
 test: lib/defs.js
-	$(MOCHA) --check-leaks -u tdd test/
+	$(MOCHA) --grep "$(TEST_FILTER)" --check-leaks -u tdd test/
 
 test-all-nodejs: lib/defs.js
 	for v in $(NODEJS_VERSIONS); \
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ function consumer(conn) {
   function on_open(err, ch) {
     if (err != null) bail(err);
     ch.assertQueue(q);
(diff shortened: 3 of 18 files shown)


hairyhum:
@squaremo this is a simpler approach to connection recovery. This approach provides a developer with examples and some helper functions to gracefully handle connection recovery without changing much code and keeping the existing objects around.

First commit allows to create consume callbacks without a channel bound in them, which can be used to re-subscribe.

Last commit fixes a bug, when error callback can be called twice. By chance it was discovered in the tests for recovery, but can be reproduced with just a simple testcase like this, which should run after `connect.js` tests:

```js
test("dummy", function(done) {
    this.timeout(15000);
    setTimeout(function(){
        done(null);
    }, 5000);
});
```

hairyhum:
The callback will be called with the actual channel in the arguments, the channel which is processing the consumed message. This will make consume function reusable because the channel argument will be the active channel.

```
function process_msg(msg, ch){
    ...
    ch.ack(msg);
}

my_channel.consume(queue, process_msg);

another_channel.consume(queue, process_msg);
```

The benefit here is that you can write a consume function separately and reuse it.
It's not necessary required for connection recovery, but makes some code much easier to write.
I can make a separate PR with that change, if you like.

squaremo:
> This will make consume function reusable because the channel argument will be the active channel.

Ah I see -- it's not critical to the recovery, just a convenience. You can also achieve this with `bind`, like

```
function process_msg(ch, msg) {
 // ...
  ch.ack(msg);
}

my_channel.consume(queue, process_msg.bind(my_queue));
another_queue.consume(queue, process_msg.bind(another_queue));
```

I am not up with the JavaScript fashions enough to know which (if either) is considered more idiomatic.

usebaz:
Hi, what is the status of this PR?

hairyhum:
@squaremo can you please update if you need any changes to that?

squaremo:
Hi, sorry, I wanted to fix the problems with NodeJS 9 and 10 (two different
problems, gah) before getting back to this -- at least so we can have
passing CI.  That's taken longer in elapsed time than I hoped. I will try
to take a look here again soon, so you at least have some feedback.

hardcodet:
@hairyhum this looks _slightly_ abandoned :). is the code here still up-to-date, or obsolete/incompatible by now?

hairyhum:
@hardcodet I didn't update this code and don't know if it's still up-to-date with the rest of the project.
In theory it should be relatively easily rebasable.

PR #436, opened 2018-06-21
Add overflow option to be set.

honestserpent:
Now when you declare a queue you can specify the overflow option (eg. "reject-publish"). ([documentation](https://www.rabbitmq.com/maxlength.html#overflow-behaviour))

Usage:
`await ch.assertQueue(q, { overflow: 'reject-publish' });`

### DIFF
--- a/lib/api_args.js
+++ b/lib/api_args.js
@@ -45,7 +45,7 @@ var EMPTY_OPTIONS = Object.freeze({});
 
 var Args = {};
 
-Args.assertQueue = function(queue, options) {
+Args.assertQueue = function (queue, options) {
   queue = queue || '';
   options = options || EMPTY_OPTIONS;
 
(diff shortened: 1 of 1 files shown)


honestserpent:
What should I do? I don't understand why checks have failed.

squaremo:
> I don't understand why checks have failed.

Me neither. Looks like a problem with npm. I've restarted the TravisCI build. If it keeps having problems, I'll have to do some investigation.

honestserpent:
@squaremo After you restart. should I do anything?

squaremo:
OK, one problem is with Node.JS 9.x, for small values of x, apparently: https://github.com/nodejs/node/issues/16781. I think I can just alter the version of Node.JS that's tested to a later one, to avoid it.

The other problem is genuinely in amqplib, just mysterious (and not your fault). It's a race in the muxing code that used to very rarely cause test flakes; now it seems to deterministically cause test flakes.

squaremo:
OK, it may take me a while to fix those CI failures. If you remove the spurious whitespace changes, I'll merge this.

honestserpent:
@squaremo pardon my ignorance but this is the first time I do a (very small) contribution on Github. Do i need to delete the Pull Request, fix it, and do it again, or is there some other way?

honestserpent:
@kibertoad thank you.
@squaremo just done.

squaremo:
> this is the first time I do a (very small) contribution on Github

@honestserpent Cool! 🎉

PR #439, opened 2018-07-09
Bump uglify version and tested NodeJS 9-series

squaremo:
Uglify 2.4 has a vulnerability that doesn't affect users of this
library, but does result in build warnings:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2015-8858

Early minor versions of NodeJS 9.x inherited a bug from V8 that led to
repeated test failures: https://github.com/nodejs/node/issues/16781.


squaremo:
NB this does not fix the failure with NodeJS 10, and some flakes may remain.

PR #447, opened 2018-07-30
Add method and class IDs to channel errors

MitMaro:
## What

This change adds a `methodId` and `classId` property to the channel errors.

## Motivation

I have a higher level configuration for the topology that allows automatic deletion of exchanges and queues. I perform a `deleteExchange`/`deleteQueue` with `ifUnused`/`ifEmpty` options set, and against exchanges and queues that no longer exist. These operations will fail with an error in some cases. Currently it is only  possible to determine if there was a `precondition­failed` error code (406) and not what operation failed. In the case of bad topology, for example a binding to a queue that does not exist, I produces a critical error fault in our monitoring system. In the case of a failure to delete an exchange or queue, I want to ignore the error, but currently cannot. With this change, errors will contain the class and method IDs, so that the source and type of error can be determined.

## Notes

* I've updated the existing tests to add checks for the class and method IDs where there was already an existing `error.code` check. I'm unsure on how or where to add any additional tests that may be needed.

### DIFF
--- a/bin/generate-defs.js
+++ b/bin/generate-defs.js
@@ -82,6 +82,8 @@ for (var i = 0, len = defs.classes.length; i < len; i++) {
     methods[name] = {
       id: methodId(clazz, method),
       name: name,
+      methodId: method.id,
+      clazzId: clazz.id,
       clazz: clazz.name,
       args: method['arguments'].map(argument),
       isReply: method.answer,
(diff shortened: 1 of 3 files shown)


MitMaro:
Fixed that use of `const` by replacing with `var`and tests are passing this time.

Also, after re-reading my previous comment about the node versions, it may have came across as being a bit rude. If so, that was not my intent. :)

MitMaro:
I also rebased onto the latest master.

PR #452, opened 2018-09-06
expose serverProperties in the connection object

jfromaniello:
Sometimes is useful to check some properties of the broker.

amqplib for python has this:
https://stackoverflow.com/a/7599072

### DIFF
--- a/test/connect.js
+++ b/test/connect.js
@@ -79,6 +79,18 @@ suite("Connect API", function() {
     connect(u, {}, kCallback(fail(done), succeed(done)));
   });
 
+  test("serverProperties", function(done) {
+    var url = require('url');
+    var parts = url.parse(URL, true);
+    var config = parts.query || {};
+    connect(config, {}, (err, connection) => {
+      if (err) { return done(err); }
+      console.dir(connection);
+      assert.equal(connection.serverProperties.product, 'RabbitMQ');
+      done();
+    });
+  });
+
   test("using custom heartbeat option", function(done) {
     var url = require('url');
     var parts = url.parse(URL, true);
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -205,6 +205,7 @@ C.open = function(allFields, openCallback0) {
                          allFields.mechanism)));
       return;
     }
+    self.serverProperties = start.fields.serverProperties;
     send(defs.ConnectionStartOk);
     wait(afterStartOk);
   }


test/connect.js
```diff
@@ -79,6 +79,18 @@ suite("Connect API", function() {
     connect(u, {}, kCallback(fail(done), succeed(done)));
   });
 
+  test("serverProperties", function(done) {
+    var url = require('url');
+    var parts = url.parse(URL, true);
+    var config = parts.query || {};
+    connect(config, {}, (err, connection) => {
+      if (err) { return done(err); }
+      console.dir(connection);
```
squaremo: Is this debug output?

test/connect.js
```diff
@@ -79,6 +79,18 @@ suite("Connect API", function() {
     connect(u, {}, kCallback(fail(done), succeed(done)));
   });
 
+  test("serverProperties", function(done) {
+    var url = require('url');
+    var parts = url.parse(URL, true);
+    var config = parts.query || {};
+    connect(config, {}, (err, connection) => {
```
squaremo: Ah -- CI is failing because this won't parse in older versions of Node.JS.

kibertoad:
@squaremo Speaking of compatibility - are there plans to ever drop support for at least the hopelessly old versions (< 4)?

squaremo:
> Speaking of compatibility - are there plans to ever drop support for at least the hopelessly old versions (< 4)?

@kibertoad It is tempting, especially as it's getting difficult to keep Node v10 happy without breaking things for those older versions. https://nodejs.org/metrics/ seems to indicate that versions before v4 are still in wide use, however.

kibertoad:
@squaremo I would argue that these numbers are pretty skewed by many popular libraries (that run CI) still supporting older versions and running tests against them; when you factor in statistics that rely on people actually providing input, numbers are very different: https://nodejs.org/en/user-survey-report/#version
Moreover, main reason why developers don't upgrade Node.js is "it's already working, don't touch anything" kind of mentality for fragile super-legacy systems. For such cases it's irrelevant whether or not there are updates for dependencies that they are using, because they are most likely not going to update anyway.

jfromaniello:
Thank you, applied your feedback:

1.  removed the console.dir I accidentally commited
2.  changed the arrow function to function

cressie176:
Re upgrading

With the increased emphasis on security from GitHub and npm (via audit), it's becoming harder and harder to support older node versions. It only takes a single dependency to introduce a newer node feature, then require a security patch, to force your hand.

PR #457, opened 2018-09-18
call .toClosed() in next tick #411

yourimiyi:


yourimiyi:
Totally agree, thanks for your attention and helps.

PR #459, opened 2018-09-21
Add david-dm dependency badges to README.md

rmharrison:
Useful for at a glance check of whether dependencies are up to date. 

NB: david and david-dm are FOSS (MIT) and non-commercial.


cressie176:
Should have been considered sooner, however since david-dm has been offline for months, I'm going to close. Thanks anyway @rmharrison

PR #469, opened 2018-10-17
Disable heartbeats if hearbeat set to 0

### LINKED ISSUE #467: Make it possible to disable heartbeats
It's currently not possible to disable heartbeats, if you set a heartbeat value of 0 on the client side it will use the server's suggested value instead of disabling it (sending 0 to the server in the TuneOk frame). 

https://github.com/squaremo/amqp.node/blob/6ee18f27e31863f5256bc921a21f36ebf2ba5e74/lib/connection.js#L192-L198

carlhoerberg:
fixes #467

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -227,7 +227,8 @@ C.open = function(allFields, openCallback0) {
       tunedOptions.channelMax =
         negotiate(fields.channelMax, allFields.channelMax);
       tunedOptions.heartbeat =
-        negotiate(fields.heartbeat, allFields.heartbeat);
+        allFields.heartbeat === 0 ?
+        0 : negotiate(fields.heartbeat, allFields.heartbeat);
       send(defs.ConnectionTuneOk);
       send(defs.ConnectionOpen);
       expect(defs.ConnectionOpenOk, onOpenOk);
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -189,7 +189,7 @@ C.open = function(allFields, openCallback0) {
     // frameMax, but we'll leave that to the server to enforce). In
     // all cases, `0` really means "no limit", or rather the highest
     // value in the encoding, e.g., unsigned short for channelMax.
-    if (server === 0 || desired === 0) {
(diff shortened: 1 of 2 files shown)


lib/connection.js
```diff
@@ -227,7 +227,8 @@ C.open = function(allFields, openCallback0) {
       tunedOptions.channelMax =
         negotiate(fields.channelMax, allFields.channelMax);
       tunedOptions.heartbeat =
-        negotiate(fields.heartbeat, allFields.heartbeat);
+        allFields.heartbeat === 0 ?
```
squaremo: If this isn't set by the user, it should accept the server value. However, it's defaulted to `0` in connect.js, so code that is currently not supplying a value will now disable heartbeats rather than using the server value.

squaremo (changes requested):

lib/connection.js
squaremo: Can this ternary expression just be simplified to

```js
tunedOptions = allFields.heartbeat === null ? fields.heartbeat : allFields.heartbeat
```

i.e., as discussed, if given in options use that value, otherwise use the server value?

carlhoerberg: ok, fixed

lib/connection.js
```diff
@@ -227,7 +227,7 @@ C.open = function(allFields, openCallback0) {
       tunedOptions.channelMax =
         negotiate(fields.channelMax, allFields.channelMax);
       tunedOptions.heartbeat =
```
ressom: I know this PR is old and @squaremo has reviewed / suggested a change, but I am thinking this is a simpler, clearer solution.
```suggestion
if (allFields.heartbeat !== 0) {
  tunedOptions.heartbeat =
    negotiate(fields.heartbeat, allFields.heartbeat);
}
```
This would also facilitate revert of the change @ line 192 above.

kibertoad:
@carlhoerberg Any chance you could finish this PR?

cressie176:
Implemented via https://github.com/amqp-node/amqplib/pull/853

PR #472, opened 2018-10-24
feat: getCh function to return channel ID

sshtel:
This getter is to return ch value for better channel resource management by users.

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -450,6 +450,9 @@ C.onBufferDrain = function() {
   this.emit('drain');
 };
 
+C.getCh = function() {
+  return this.ch;
+}
 
 // This adds just a bit more stuff useful for the APIs, but not
 // low-level machinery.


squaremo:
> better channel resource management

Can you elaborate on how this helps with better channel resource management?

sshtel:
When it comes to customized channel pooling, we need to know channel's ID to choose if the channel should be released or reused.

Here, we are using a simple array for channel pooling.
https://github.com/spearhead-ea/island/blob/f880180a8839bf0f998a2b512e5775f691789fc5/src/services/amqp-channel-pool-service.ts#L23
`private idleChannels: Promise<amqp.Channel>[] = [];`

When I need to get rid of unused channel from the pool in 'onClose' handler, there is no way to know which channel should be removed.

amqp.node provides good channel management APIs but still we need more flexible resource management solution to handle a bunch of consumer's request.

https://github.com/spearhead-ea/island/blob/f880180a8839bf0f998a2b512e5775f691789fc5/src/services/amqp-channel-pool-service.ts#L121


Plus,
It would be great if `amqp.node` provides embedded channel pool feature in the future.

squaremo:
That's a really interesting use! Is object equality not enough for knowing which channel it is?
The channel ID may be an unreliable identifier, because the IDs themselves are reused (although not while a channel is still open).

sshtel:
Comparing object equality was working before I changed the array to Promise array.
WAS: `private idleChannels: amqp.Channel[] = [];`
AS-IS: `private idleChannels: Promise<amqp.Channel>[] = [];`

Promise objects cannot be compared because its internal event count is changing every time.. So I had to use channel ID to know the same object to purge.

I didn't know that channel ID is reused internally. But, yes, it will not that big problem if it is unique when the channel is alive.

sshtel:
For remind, any thoughts for this MR? @squaremo

cressie176:
I have to do something similar to this in [Rascal](https://github.com/guidesmiths/rascal) in order to support channel pools, but rather than use the actual channel id I [decorate the channel object with my own](https://github.com/guidesmiths/rascal/blob/master/lib/amqp/Vhost.js#L404). I think this approach would be safer than using the channel number, which as squaremo said my be reused.

Any objections to adding a uuid property to the channel instead of exposing the underlying channel number?

cressie176:
I've been looking at this again, and have a couple of issues with it...

1. channel.ch is already public, and not prefixed by an underscore, so hiding it behind a function doesn't add very much
2. I believe the channel id only unique per connection, so may not be unique enough.

cressie176:
Closing as the channel id is already accessible

PR #475, opened 2018-10-28
Prep v0.5.3

cressie176:
The 0.8 tests failed with
```
  1) Connect API wrongly typed heartbeat option:
     Error: done() called multiple times

  2) Connect API wrongly typed open option:
     Error: done() called multiple times
```

`nave` prompted for a version when it came to iojs, so I just hit enter

node 10 had the following deprecation warnings
```
(node:41493) [DEP0006] DeprecationWarning: child_process: options.customFds option is deprecated. Use options.stdio instead.

(node:41494) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
  
(node:41494) [DEP0097] DeprecationWarning: Using a domain property in MakeCallback is deprecated. Use the async_context variant of MakeCallback or the AsyncResource class instead.
```

I'm not sure which of these are something to worry about. Took a quick look at the 0.8 issue - it only manifests when all the tests are run.


squaremo:
> The 0.8 tests failed with

I suspect there's a single cause to those two failures. Will have a look tonight.
There's agitation for dropping support for older Node.JS versions, but I don't want to do it just to avoid having to face bugs.

> `nave` prompted for a version when it came to iojs

Oh yeah, iojs -- is that a thing any more? (in the same way as e.g., Node.JS 0.8 is) -- or did they all go away.

> node 10 had the following deprecation warnings

Those must be from libraries; I guess we could update the dependencies at some point to see if that gets rid of them.

squaremo:
I've fixed the repeating failures, but there are still occasional flakes. I think it's good enough to crack on though.

squaremo:
> (was a bit busy last week - just started an MBA)

Cool! Yes that'll keep you busy :-)

cressie176:
@squaremo - a couple more changes...

1. I merged your dependency changes to master

1. I'd already added Node 11.1 to the make file, but neglected to update .travis. I'd rather recognise contributors, so instead of just updating .travis myself I merged @kibertoad's [PR](https://github.com/squaremo/amqp.node/pull/473)

I had a random couple Travis test failure in Node 10 (done called twice), but they worked when I re-ran. Assuming the tests are a little flakey?

Let me know when/if you're happy that this is good to go.

squaremo:
Just removing another deprecation message by updating bitsyntax-js (commit by commit! oof).

cressie176:
Good to go?

squaremo:
Board is green -- I reckon go for it!

PR #477, opened 2018-10-31
Update README promise example

renarsvilnis:
Hey,
I started learning Rabbitmq in node using this library and noticed that the readme promise example wasn't clear enough and didn't follow current node.js tendencies.

I updated the example, now It's way more readable and easier to understand what is going on for someone starting out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ function consumer(conn) {
   function on_open(err, ch) {
     if (err != null) bail(err);
     ch.assertQueue(q);
-    ch.consume(q, function(msg) {
+    ch.consume(q, (msg) => {
       if (msg !== null) {
         console.log(msg.content.toString());
         ch.ack(msg);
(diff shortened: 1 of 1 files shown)


squaremo:
Thank you for this update to the example -- I guess it is past time to adopt more modern JavaScript syntax.

README.md
```diff
@@ -74,35 +74,33 @@ require('amqplib/callback_api')
...
-      if (msg !== null) {
-        console.log(msg.content.toString());
-        ch.ack(msg);
-      }
-    });
-  });
-}).catch(console.warn);
+  // Sender
+  setInterval(async () => {
+    await ch.sendToQueue(queue, Buffer.from('something to do'));
```
squaremo: No need to `async/await` here, the return value is not a promise

renarsvilnis:
Hi @squaremo,
Add the changes you wanted, also simplified other examples by moving to arrow function.

kslr:
Why is there no progress?

PR #479, opened 2018-11-08
assertQueue - allow additional option to be passed while declaring: p…

lukiq:
…assive


squaremo:
This is what `checkQueue` does. Since it changes the meaning of all the other parameters (it ignores them), I made it another method.

squaremo:
@lukiq Are you able to do what you needed by using `checkQueue` instead of `assertQueue({passive: true})`?

squaremo:
I'm going to close this on the presumption that checkQueue does what was needed. Thank you for taking the time to put in a PR though!

PR #498, opened 2019-03-24
Error pending confirmation callbacks on channel close

johanneswuerbach:
Otherwise waitForConfirms never terminates or terminates node

Additional this includes a commit https://github.com/squaremo/amqp.node/commit/099673edefdc3a7b570027c6f48793bbecf5789b, which makes it possible to install this module directly from git (instead of the published package), but I'm happy to drop this if preferred.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,6 +4,5 @@ scratch
 # node_modules is ignored anyway
 .travis.yml
 bin/amqp-rabbitmq-0.9.1.json
-bin/generate-defs.js
 etc/
 coverage/
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -33,6 +33,12 @@ function Channel(connection) {
   this.on('nack', this.handleConfirm.bind(this, function(cb) {
     if (cb) cb(new Error('message nacked'));
   }));
+  this.on('close', function () {
+    var cb;
+    while (cb = this.unconfirmed.shift()) {
+      if (cb) cb(new Error('channel closed'));
+    }
+  })
   // message frame state machine
   this.handleMessage = acceptDeliveryOrReturn;
 }
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "scripts": {
     "test": "make test",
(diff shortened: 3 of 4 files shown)


squaremo:
This makes sense. It is a behaviour change though -- people may not be prepared for rejections from waitForConfirms (though I hope they would be). I guess we can call that out in release notes when it comes to it.

The tests are failing because of some ES6 syntax which snuck in channel.js L36 (honestly, if I had the time I'd rewrite in ES6 and babel it for older runtimes ..).

johanneswuerbach:
While I understand the behavior change I think the current behavior is worse as messages won’t be acknowledged (the channel is closed) so the promise will never resolve and leave the process just hanging, which seems more like a bug.

Afaik you already need to handle the reject case anyway as messages could be rejected for other reasons, not?

Will fix the es6 syntax, which node versions are currently supported?

squaremo:
> Afaik you already need to handle the reject case anyway as messages could be rejected for other reasons, not?

Absolutely -- in this case the risk is people relying too much on the "happy path", so I'm not too troubled by it.

> Will fix the es6 syntax, which node versions are currently supported?

All the way back to Node 0.8. (!)

johanneswuerbach:
@squaremo tests fixed.

johanneswuerbach:
@squaremo friendly ping.

johanneswuerbach:
Maybe @cressie176? Sorry for the spam, but installing this dependency from git requires curl, which is a bit of a bummer :-(

johanneswuerbach:
@squaremo would you also be able to have a look at https://github.com/squaremo/amqp.node/pull/503?

squaremo:
Sorry @johanneswuerbach, I don't always get a lot of time to react to PRs and issues -- but at least you've made it easy for me by explaining the motivation and keeping the change simple

johanneswuerbach:
@squaremo sure, no worries. As an open source maintainer I completely understand, let me know if there is anything I can do to help.

PR #503, opened 2019-04-01
Fix stream becoming readable again

johanneswuerbach:
I'm not really sure how to test this for real, but under constant publishing I can trigger an occasional `Uncaught AssertionError [ERR_ASSERTION]: 0 == 1` https://github.com/squaremo/amqp.node/blob/c28c176dd74565c73957eebf00932a892d2aa840/lib/mux.js#L92

It seems that some kind of stream in this library is emitting `readable`, during read calls and causing this assertion to trip.

### DIFF
--- a/lib/mux.js
+++ b/lib/mux.js
@@ -55,26 +55,11 @@ Mux.prototype._readIncoming = function() {
   // are empty, or we exhaust our ability to accept chunks.
   function roundrobin(streams) {
     var s;
-    // if there's just one incoming stream we don't have to
-    // go through all the dequeue/enqueueing
-    if (streams.length === 1) {
-      s = streams.shift();
-      while (accepting) {
-        var chunk = s.read();
-        if (chunk !== null) {
-          accepting = out.write(chunk);
-        }
-        else break;
-      }
-      if (!accepting) streams.push(s);
-    }
-    else {
-      while (accepting && (s = streams.shift())) {
-        var chunk = s.read();
-        if (chunk !== null) {
-          accepting = out.write(chunk);
-          streams.push(s);
-        }
+    while (accepting && (s = streams.shift())) {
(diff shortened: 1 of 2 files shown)


squaremo:
I think I understand what the problem is; I need to think more about the solution though

johanneswuerbach:
@squaremo sorry for the ping, but do you have any update? :-)

johanneswuerbach:
@squaremo anything I could do to help moving this forward?

squaremo:
Right I see, the problem is that (as you say) a stream can be added back into `newStreams` after the initial loop exits, and before the assertion that `newStreams` is empty. The loop exits because a null chunk has been read (indicating no more chunks); but, 'readable' fires (synchronously, somehow) and the handler for that event appends the stream to `newStreams`. This breaks the invariant that either all `newStreams` are exhausted (and removed from that queue), or the downstream is full.

Your fix restores the invariant by putting another loop around the problematic loop, to catch the case where the loop exits but the stream has been added back by the 'readable' event handler.

I think another way to fix it would be to simply remove the special case for when there's a single stream. Since that loop will only exit when `!accepting` (as before) or the queue is empty, it will have the same effect as the extra loop. It's possible, but harmless, for a stream to be added back to the queue twice.

johanneswuerbach:
Sorry for the late response @squaremo, but now updated the PR with your suggestion and it still works for us 👍

johanneswuerbach:
@squaremo sorry for the ping, but is there anything else required here?

johanneswuerbach:
@squaremo thank you for the support. Could you also release a new version with this fix? That would be wonderful.

axnsan12:
@squaremo can you please publish a release with this fix?

glebsts:
Why it had no changelog update?

PR #514, opened 2019-05-02
Close and destroy sockets on connection error.

### LINKED ISSUE #513: Failed "connect" call does not close socket
I am developing integration tests and have problems with negative case testing - my test app never exits.

Steps to Reproduce:
* Call connect to vhost, for which user does not have permissions
* See that err callback is called
* See that NodeJS app just hangs, never exits.

I found that reason is that the following block is only called on timeout:
```
        sock.end();
        sock.destroy();
```
And it is never called on other connection errors.

rumata28:
Add `sock.end` and `sock.destroy` calls for `sock.once('error', ...)`
event handler and for `c.open(fields, ...)` callback - just like it was
done for connection timeout callback.

Resolves #513

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -151,7 +151,11 @@ function connect(url, socketOptions, openCallback) {
       if (err === null) {
         openCallback(null, c);
       }
-      else openCallback(err);
+      else {
+        sock.end();
+        sock.destroy();
+        openCallback(err);
+      }
     });
   }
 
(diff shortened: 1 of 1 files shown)


rumata28:
Sorry, tests failed.
I use Windows as a development machine, need to figure it out how to install everything.
Just checked that the change fixes issue for my tests, locally.

rumata28:
Close it, will reopen once I make it work.

PR #525, opened 2019-07-12
Update package.json to reflect supported node versions

tingwai:
Based on the current `.travis.yml`, node version 10, 11, 12 are supported. I have updated the package.json to reflect that. Otherwise amqplib fails to install when users have flag `engine-strict=true` in their .npmrc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/squaremo/amqp.node.git"
   },
   "engines": {
-    "node": ">=0.8 <=9"
+    "node": ">=0.8 <=12"
   },
   "dependencies": {
     "bitsyntax": "~0.1.0",


package.json
```diff
@@ -9,7 +9,7 @@
     "url": "https://github.com/squaremo/amqp.node.git"
   },
   "engines": {
-    "node": ">=0.8 <=9"
+    "node": ">=0.8 <=11"
```
L2jLiga: I guess it should be `<= 12`

tingwai: Ah you're right, updating now

squaremo:
You'll need to rebase on master, which has a fix for the CI build (see #526).

tingwai:
> You'll need to rebase on master, which has a fix for the CI build (see #526).

Thanks for the heads up. PR is rebased now

squaremo:
Thanks @tingwai (and @L2jLiga for the assist)!

tingwai:
@squaremo Thanks for the merge. Can you release a new version to npm? This will help resolve the engine strict issues in my CI

PR #528, opened 2019-07-19
Update the deprecated module with new module

StrayBird-ATSH:
The node-uuid module is deprecated and should not be used any more. Change this dependency to uuid module is appropriate.


StrayBird-ATSH:
No. Since the uuid module has the backward compatibility ability, no other change is needed. 
Thanks for your review.

PR #531, opened 2019-07-23
Introduce post-connection timeout setting

hardcodet:
Currently, the timeout for the underlying socket is reset to `0` after a successful connection. This can lead to promises hanging indefinitely / callbacks not being invoked in case the server becomes unresponsive. On the other hand, just using the default socket timeout may be too short, and introduce runtime errors.

As a solution, this change introduces a second socket option `postConnectTimeout` which is being applied to the socket after a successful connection, rather than resetting the timeout to `0`. If the setting is not provided, a high default timeout of 300000 (5 minutes) is being set, which would still prevent indefinitely blocking clients.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -96,6 +96,7 @@ function connect(url, socketOptions, openCallback) {
 
   var noDelay = !!sockopts.noDelay;
   var timeout = sockopts.timeout;
+  var postConnectTimeout = sockopts.postConnectTimeout || 300000
   var keepAlive = !!sockopts.keepAlive;
   // 0 is default for node
   var keepAliveDelay = sockopts.keepAliveDelay || 0;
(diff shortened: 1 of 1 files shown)


cressie176:
This looks like a good idea to me, although I think unnecessary if heartbeats or TCP Keep Alive is enabled. Since heartbeats are enabled by default, I would change the default post-connection timeout to 0, but still support it as an option. Thoughts @hardcodet, @kibertoad?

cressie176:
I've reviewed again. The change is more complicated than it first appears. 

Setting a new socket timeout value does not clear the pre-connection timeout listener. We should do this at the same time as setting the timeout to zero, then set a post connection timeout if configured. The post connection timeout listener should destroy the socket passing it an error so that the a connection error is emitted

cressie176:
I'm not confident to make this change.

PR #536, opened 2019-07-30
fix: travis ci script to use xenial VM

sshtel:
- use xenial VM image for travis CI.
- ~apt-get update before_install~
- ~npm run prepare to build before deploy~

since 0.5.5 is released, commit is updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: node_js
 node_js:
 - "0.8"


sshtel:
image

To fix apt get install error, 

xenial VM is used

squaremo:
I think the default is xenial, isn't it? That's why I had to redo a bit of the config, to install rabbitmq.

sshtel:
Not sure if the default is xenial, I tried travis CI in my folk repo and failed multiple time until I implicitly define `dist`

if this is not needed, I will close.

sshtel:
https://docs.travis-ci.com/user/reference/overview/

Trusty is default yet maybe?

sshtel:
Since there is no issue for now, I will close this.

PR #556, opened 2020-02-11
Fix Heartbeat Intervals

seesemichaelj:
Hello!

I'm using this library in a production setting; and I noticed that under load I would occasionally timeout due to missing heartbeats. Looking at the logic of determining whether or not to send a heartbeat, I realized the library actually doesn't send heartbeats at t/2 intervals (as intended by the comments in `lib/heartbeat.js`) but t intervals. Under load, we could miss timeouts.

This PR changes the way the library determines whether or not we should send a heartbeat by using time with a buffer rather than a boolean flag. You'll see a test I added in https://github.com/squaremo/amqp.node/commit/fd47fbdb8485843307feb7a82c6b2d6c61c61035 that demonstrates the bug (by failing). 

Here's a potential scenario of current behavior that is failing in my production environment:
1. `sendHearbeat` is called
1. `sendBytes` is called inside `sendHeartbeat`
1. `setSinceLastCheck` set to `true` inside of `sendBytes`
1. No network traffic occurs
1. `intervalMs / 2` time elapses and `Heart.runHeartbeat` is called
1. `runHeartbeat` runs `check` which is `Connection.checkSend` in this case
1. `checkSend` returns the old value of `setSinceLastCheck` (which is `true`) and sets `setSinceLastCheck` to `false`
1. `runHeartbeat` doesn't `fail()` due to `!true` condition and therefore doesn't send a heartbeat when we'd expect it to send a heartbeat every `intervalMs / 2` given no other traffic
1. No network traffic occurs
1. `intervalMs / 2` time elapses. We're now very close to `intervalMs` which is the timeout period. Under load, the `setInterval` may be a little late and miss the timeout period
1. We get a heartbeat timeout from the server because we sent a beat a tad late

I've tested these changes in my production environment, and seem to work great

Let me know if you have any questions!

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -520,12 +520,19 @@ C.checkRecv = function(intervalMs) {
 
 C.sendBytes = function(bytes) {
   this.timeLastSent = Date.now();
-  this.stream.write(bytes);
+  return this.stream.write(bytes);
 };
 
 C.sendHeartbeat = function() {
   console.log("Sending heartbeat");
-  return this.sendBytes(frame.HEARTBEAT_BUF);
+  var flushedBuffer = this.sendBytes(frame.HEARTBEAT_BUF);
+
+  if (!flushedBuffer) {
+    console.log("listening for drain");
+    this.socket.once("drain", console.log("drained"));
+  }
+
+  return flushedBuffer;
 };
 
 var encodeMethod = defs.encodeMethod;
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -15,7 +15,8 @@ var Duplex =
   require('stream').Duplex ||
   require('readable-stream/duplex');
 var EventEmitter = require('events').EventEmitter;
-var Heart = require('./heartbeat').Heart;
+var heartbeat = require('./heartbeat');
+var Heart = heartbeat.Heart;
 
 var methodName = require('./format').methodName;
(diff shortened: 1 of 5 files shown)


cressie176:
Thanks for the PR, and sorry it has taken so long to respond. I recently investigated https://github.com/amqp-node/amqplib/issues/588#issuecomment-1126902872 and found that RabbitMQ doesn't close the connection until long after the heartbeat (or any other command) is overdue. 

Consequently I'm going to close this, but happy to reconsider if I'm wrong about RabbitMQ's behaviour.

seesemichaelj:
While I'm no longer using this library (no longer at that company and the product was also discontinued), we did get our connections closed because of missed heartbeats. This PR was necessary for us to operate in high load scenarios. I no longer have a stake in this, but there is an implementation issue as it stands.

cressie176:
Understood, and thanks for the update. I had the same concerns as you, but my tests clearly show it waits for almost 3x the heartbeat timeout, which is enough for one dropped heartbeat and a very slow one. It's possible I used a more modern version RabbitMQ, updated allow a longer grace period. I'll keep an eye out for any more reports.

seesemichaelj:
Was actually about to update my comment: it's definitely possible the observed behavior was different when this PR was made. Thanks anyway for the response (and for the library; it served us well); have a good one!

PR #566, opened 2020-03-28
Add example for async/await in README

simone-sanfratello:
Hello,

it may be useful an example in ES6+ js style

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,42 @@ open.then(function(conn) {
 }).catch(console.warn);
 ```
 
+## async/await API example
+
+```javascript
+const amqp = require('amqplib')
+
+async function publisher ({ host, queue }) {
+  try {
+    const connection = await amqp.connect(host)
+    const channel = await connection.createChannel()
+    await channel.assertQueue(queue)
+    await channel.sendToQueue(queue, Buffer.from('something'))
+  } catch (error) {
+    console.warn(error)
+  }
+}
+
+async function consumer ({ host, queue }) {
+  try {
+    const connection = await amqp.connect(host)
+    const channel = await connection.createChannel()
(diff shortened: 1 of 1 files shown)


fatihayyildiz:
@simone-sanfratello Hello Simone,

Thanks for the PR but when I tried this implementation at node.js 12.16.2 I am getting error. 
This line is not creation connection variable and one line after it, node.js complain about connection variable because it comes undefined. 

```javascript
const connection = await amqp.connect(host)
```

Error message : TypeError: Cannot read property 'createChannel' of undefined. I also want to implement with ES6+ js style but it seems not working. Any idea about this issue ? 

BTW: my implementation is like this and I am running this file with "node app.js" command.

```javascript
const amqp = require('amqplib/callback_api');

async function consumer ({ host, queue }) {
  try {
    console.log('host :',host)
    const connection = await amqp.connect(host)
    console.log('connection :',connection ) // <-- printing undefined
    const channel = await connection.createChannel()
    await channel.assertQueue(queue)
    console.log('consumer :',queue)
    channel.consume(queue, (msg) => {
      if (msg !== null) {
        console.log('received', msg.content.toString())
        channel.ack(msg)
      }
    })
  } catch (error) {
    
    console.warn(error)
    bail(error)
  }
}
consumer({ host: "amqp://"+process.env.RABBITMQ_URL, queue: 'content_queue' });

```

simone-sanfratello:
@fatihayyildiz 

modify the first line in 

```js
const amqp = require('amqplib')
```

fatihayyildiz:
> @fatihayyildiz
> 
> modify the first line in
> 
> ```js
> const amqp = require('amqplib')
> ```

Yes it working now @simone-sanfratello . I miss that first line. Thanks for your help

PR #567, opened 2020-03-31
set SNI hostname for TLS connections

carlhoerberg:
Is required if you have multiple AMQP servers behind a TLS load balancer.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -106,6 +106,7 @@ function connect(url, socketOptions, openCallback) {
   if (typeof url === 'object') {
     protocol = (url.protocol || 'amqp') + ':';
     sockopts.host = url.hostname;
+    sockopts.servername = url.hostname;
     sockopts.port = url.port || ((protocol === 'amqp:') ? 5672 : 5671);
 
     var user, pass;
@@ -130,6 +131,7 @@ function connect(url, socketOptions, openCallback) {
     var parts = URL(url, true); // yes, parse the query string
     protocol = parts.protocol;
     sockopts.host = parts.hostname;
+    sockopts.servername = parts.hostname;
     sockopts.port = parseInt(parts.port) || ((protocol === 'amqp:') ? 5672 : 5671);
     var vhost = parts.pathname ? parts.pathname.substr(1) : null;
     fields = openFrames(vhost, parts.query, sockopts.credentials || credentialsFromUrl(parts), extraClientProperties);


sfsr12:
This should really get merged.  I just spent 3 hours trying to figure out what was wrong with my TLS load balancer, and I finally found this when I realized it wasn't an issue on the termination end.

danwatford:
Mine is a similar story to @sfsr12 , I had been trying to connect to RabbitMQ behind a Traefik reverse proxy using LetsEncrypt certificates for several hours

amqps connections fail with the following error:
```
Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1497:34)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12) {
  code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
```

Amqplib's SSL guide links to [node's TLS documentation](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback) where SNI is mentioned:

> Unlike the https API, tls.connect() does not enable the SNI (Server Name Indication) extension by default, which may cause some servers to return an incorrect certificate or reject the connection altogether. To enable SNI, set the servername option in addition to host.

Solution is to specify the servername is the socketOptions object passed to amqplib's connection function, however merging this PR is preferred. Or possibly an update to the amqplib SSL Guide's list of errors explaining that use of a reverse-proxy may necessitate setting the servername option.

djmitche:
Agreed, this just caused a rollback in a maintenance window for us.

Is there interest in a forked version of amqplib that does include this PR?

squaremo:
OK yes, I see the requirement. What will happen if someone uses an IP address as the hostname?

djmitche:
For TLS, I don't think that will work, with or without `servername`.

The `servername` is only used by the server to select a certificate when there are multiple certificates available.  If someone has the current version of amqplib using TLS with an IP address, then the server must be returning a certificate with that IP as its CN. This is not sometihng that publicly trusted CAs would do, so this would be a self-signed cert, and is likely the only certificate the server has available -- thus it would either ignore the `servername` or find that the `servername` matches the certificate CN, and select that certificate as it had before.

squaremo (approved):
Seems like this will avoid some pain with figuring TLS out -- I'm all for avoiding pain :+1: Thank you Carl!

Issue #549, opened 2019-12-09
node version 13


wormen:
![image](https://user-images.githubusercontent.com/8462339/70441893-8ed80000-1ad8-11ea-9204-2e9ac02e7f44.png)

Maxpain177:
@squaremo please fix it

PR #570, opened 2020-05-04
Allow newer node version

w1gz:
Fixes https://github.com/squaremo/amqp.node/issues/549.

# Context

We are using `amqp.node` with the latest NodeJS version (i.e. 13 and 14) but unfortunately,
`amqp.node` has its `package.json` fixed to an older version of NodeJS (12).

# Fix

This PR has two commits:

 - One for allowing newer engines of NodeJS
 - The other to support those new engines in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,9 @@ node_js:
 - "9.11"
 - "10.10"
 - "11.1"
-- "node"
+- "12"
+- "13"
+- "node" # Latest stable Node version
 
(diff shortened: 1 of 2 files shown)


yekurtal:
@squaremo Can we get this merged? Lots of people forking, which can cause security issues when forks are not updated regularly.

squaremo:
Will need to do a release as well, stand by ..

Issue #418, opened 2018-05-03
Additional maintainers


kibertoad:
Given that this project is very quiet lately and it doesn't look like anyone is addressing PRs that get piled up, it would probably be most helpful if there were additional maintainers with write permissions who could review, approve and merge PRs.

Stradivario:
i agree

yaronlevi:
So this project is not actively maintained

cressie176:
> So this project is not actively maintained?

I think that depends on what you mean by active. The AMQP 0-9-1 protocol was in published in November 2008, so this module doesn't have much need to be updated frequently. Despite this version 0.5.2 was released in November 2017. There's a reasonable response rate to issues. Given the amqp.node protocol is complicated and bug ridden I can understand some reluctance to add more maintainers with write permissions.

kibertoad:
@cressie176 However, when community creates something useful like #411, it is really unfortunate when there is noone around to review it. Don't get me wrong, I am very grateful to the author for all the hard work he has been putting into the project over the years, all I am worried about is project stagnating due to lack of gatekeepers around.

squaremo:
> Given that this project is very quiet lately and it doesn't look like anyone is addressing PRs that get piled up, it would probably be most helpful if there were additional maintainers with write permissions who could review, approve and merge PRs.

I think that's a fair statement. To clear thing up: I have not abandoned amqplib, I just have very little time to spend on it.

It would be good to have additional maintainers. How shall I pick them?

michaelklishin:
@squaremo team RabbitMQ tried to get in touch and ask for commit rights for this project for at least a couple of us. Some users of this client are much more active in Github issues than others @cressie176 comes to mind.

I don't know if "several weeks" is a reasonable response rate. If my response rate for a project is weeks I always add most active users as committers. This client badly needs some quality time (like no other) and team RabbitMQ is happy to offer some. We are not JavaScript experts but we have more experience maintaining RabbitMQ clients than any other team on this planet.

michaelklishin:
Besides the core team this project badly needs an issue cleanup. Most of the open issues are questions, some are basic, that should be closed, labelled as such and answered if there's enough information to work with. The rest will be triaged to see if they are legitimate and can be addressed.

At some point maintainers will have to produce new releases, which means they must have push permissions on NPM and potentially more services.

cressie176:
[...]

PR #584, opened 2020-08-21
Close and destroy underlying socket after any connection error

### LINKED ISSUE #513: Failed "connect" call does not close socket

xabinapal:
Ensure that on connection errors (either from the underlying socket or from the AMQP protocol) no more messages are send to the destination and fully close and destroy the socket. This fixes the open socket leak after sending invalid connection options, such as invalid credentials or virtual host, for example.

Based on #514, resolves #513.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -150,8 +150,11 @@ function connect(url, socketOptions, openCallback) {
       if (timeout) sock.setTimeout(0);
       if (err === null) {
         openCallback(null, c);
+      } else {
+        sock.end();
+        sock.destroy();
+        openCallback(err);
       }
-      else openCallback(err);
     });
   }
 
@@ -174,7 +177,11 @@ function connect(url, socketOptions, openCallback) {
   }
 
   sock.once('error', function(err) {
-    if (!sockok) openCallback(err);
+    if (!sockok) {
+      sock.end();
+      sock.destroy();
+      openCallback(err);
+    }
   });
 
 }
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -179,8 +179,15 @@ C.open = function(allFields, openCallback0) {
   function send(Method) {
     // This can throw an exception if there's some problem with the
     // options; e.g., something is a string instead of a number.
-    try { self.sendMethod(0, Method, tunedOptions); }
-    catch (err) { bail(err); }
+    try {
+      self.sendMethod(0, Method, tunedOptions);
+    } catch (err) {
+      bail(err);
+
+      // We are in an inconsistent state so we can't continue,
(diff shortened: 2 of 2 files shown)


carlhoerberg:
Thank you! Works great

carlhoerberg:
```sh
npm install squaremo/amqp.node#pull/584/head
```
for those of you that want to use this PR in your apps

xabinapal:
Didn't know that installing versions from PRs was an option, and this not being merged was a stopper for me. Good to know!

Just asking, @squaremo is this project not active anymore?

xamgore:
@xabinapal from your perspective, is it possible to cover the code with tests? It would allow @squaremo to merge the PR without hesitation.

xabinapal:
I haven't thoroughly investigated how tests work in this library, only to ensure all test cases continue working as expected. I think it would be technically possible to add some tests to check that connection closes after an error.

Anyways, this library seems to be 100% abandoned. So many issues and PRs without owner's feedback, and this #418 issue searching for maintainers is also quiet since 2018.

With zero confidence on this PR being merged, I'm not willing to spend any time improving it, sorry. I'have been using it in production thanks to https://github.com/squaremo/amqp.node/pull/584#issuecomment-701591435 with no problems, and I encourage everyone who needs this to use that method. If @squaremo or any other maintainer relaunches this project, I will gladly spend more time on this issue.

xamgore:
@cressie176 maybe you could become a new maintainer and help @squaremo with the project?

kibertoad:
@xabinapal I know that quite a bit of time has passed, but would you be open to updating this PR? Library is not quite dead these days :)

cressie176:
The original issue (#513) identified two potential socket leak paths in lib/connect.js. 

The first occurs when c.open() completes with an error - for example, when the server refuses the connection due to bad credentials or an invalid vhost. In this case the TCP connection was successfully established (sockok = true) but the underlying socket was never explicitly closed, leaving it open. This was fixed in PR #647, which added sock.end() and sock.destroy() in the c.open() error callback. 

The second path is the sock.once('error', ...) handler, which is guarded by !sockok - meaning it only fires if the socket emits an error before the TCP connection completes (e.g. ECONNREFUSED). In that scenario the socket never successfully opened and is already dead by the time the error fires, so calling destroy() would be a no-op. There is no socket leak in this path and no fix is required.

PR #596, opened 2020-11-26
Add option to set the passive flag for assertQueue & assertExchange

dcermak:
This allows users to set the passive flag when using `assertQueue` and `assertExchange`, as certain rabbitmq servers require that and will refuse connections otherwise (e.g. rabbit.opensuse.org)

### DIFF
--- a/lib/api_args.js
+++ b/lib/api_args.js
@@ -67,7 +67,7 @@ Args.assertQueue = function(queue, options) {
     durable: (options.durable === undefined) ? true : options.durable,
     autoDelete: !!options.autoDelete,
     arguments: argt,
-    passive: false,
+    passive: !!options.passive,
     // deprecated but we have to include it
     ticket: 0,
     nowait: false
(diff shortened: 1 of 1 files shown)


squaremo:
Thanks for posting this PR ✋ The `checkQueue` and `checkExchange` methods are like `assertQueue` and `assertExchange`, but with the `passive` flag. I made them different methods because when `passive` is set, the behaviour is somewhat different. Using `checkExchange` with the openSUSE server should work.

dcermak:
> Thanks for posting this PR hand The `checkQueue` and `checkExchange` methods are like `assertQueue` and `assertExchange`, but with the `passive` flag. I made them different methods because when `passive` is set, the behaviour is somewhat different. Using `checkExchange` with the openSUSE server should work.

Actually it does not with `checkQueue` and `checkExchange`, but with `assertQueue` and `checkExchange`. I thought that I tried all possible combinations last year before I submitted this PR, but apparently I did not try them all.

Issue #265, opened 2016-08-13
Allow overriding Promise library


NotBobTheBuilder:
Currently the `when` Promise library is hardcoded. A useful feature might be to allow overriding this when requiring, in case the implementer has a preference for another promise library.

For example, the `pg-promise` library achieves this with the pattern: `require('pg-promise')({promiseLib: require('bluebird')})`

p-bakker:
Also an issue with the dependency on Bluebird, as Bluebird doesn't support async_hooks in NodeJS, which results in amqplib not supporting async_hooks

xenoterracide:
really want this so we only need native promises, would a PR make it happen?

Worthaboutapig:
I've done this and created a PR, though with a few caveats which shouldn't be a problem for most: https://github.com/squaremo/amqp.node/pull/599

cressie176:
I'm going to close this in favour of just using native promises https://github.com/amqp-node/amqplib/issues/627. 

Thanks for the PR @Worthaboutapig, and sorry it didn't get reviewed. I took a quick look just now and it includes a few unrelated changes (typescript defs, linting, plus maybe some dependency bumps), so I wouldn't want to merge. It will be useful as a reference point though.

PR #599, opened 2021-01-13
Allow overriding Promise library #265

Worthaboutapig:
This is done, however, there are some failing test on apparently unrelated stream issues.
There's a few changes to make tests etc easier to run under an IDE.
(Updated packages and uses pnpm)


cressie176:
Thank you for the PR and sorry it has taken so long to respond. I'm closing in favour of native promise support now that support for older node versions has been dropped.

Worthaboutapig:
> Thank you for the PR and sorry it has taken so long to respond. I'm closing in favour of native promise support now that support for older node versions has been dropped.

No problem. The approach I took handled this by using the native Promise library by default, but allowing people to pass in their own if they liked. This seems the most flexible option.

Is there an issue for the changes, are they in progress? My PR works fine in practice (though appreciate isn't necessarily suitable for merging).

cressie176:
Hi @Worthaboutapig, I thought about whether we want people to be able to supply their own promise library and decided against it. I agree it would be more flexible, but the internals of the library should be opaque to users, and it would create potential for some difficult to diagnose bugs if we allowed people to specify their own.

https://github.com/amqp-node/amqplib/issues/627 exists and there is also a PR https://github.com/amqp-node/amqplib/pull/624

At the moment I'm picking off some smaller/simpler changes though.

Worthaboutapig:
Ok, thanks. I made the changes because Bluebird created a bunch of problems in debugging, as it throws loads of exceptions as part of its normal startup behaviour. It was horrendous to the point of making my whole debugging experience with RabbitMQ based stuff unusable- so IMHO it's critical to being useful.
Perhaps others don't have the same debugging experience I do though.

PR #605, opened 2021-02-17
Bump NPM packages and Node version allowed

### LINKED ISSUE #593: node v15 support
Why 
```
"engines": {
  "node": ">=0.8 <=14"
}
```
I dont see any restrictions for v15

squaremo:
Fixes #593.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/squaremo/amqp.node.git"
   },
   "engines": {
-    "node": ">=0.8 <=14"
+    "node": ">=0.8 <=15"
   },
   "dependencies": {
     "bitsyntax": "~0.1.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "url": "https://github.com/squaremo/amqp.node.git"
   },
   "engines": {
-    "node": ">=0.8 <=14"
+    "node": ">=0.8 <=15",
+    "npm": "<= 6 >= 7.0.8"
```
kibertoad, line 13: Is this correct?

squaremo: That's what was suggested in https://github.com/squaremo/amqp.node/issues/593#issuecomment-729434507

squaremo: @kibertoad Do you know it to be incorrect, or did it just stand out as looking odd or incongruous?

squaremo: I removed that line, as it's not directly relevant. It can be fixed in another PR.

kibertoad: Official documentation (https://docs.npmjs.com/cli/v7/configuring-npm/package-json) lists `>1.0.2 <=2.3.4` as an example, but probably it also works in reverse as well. Mostly I was confused why npm 6 is not supported, but I guess #593 answers that.

PR #606, opened 2021-02-25
update url-parse from ~1.4.3 to ~1.5.1

ThomasGawlitza:
According to https://snyk.io/vuln/npm:url-parse, version 1.5.1 has no known direct vulnerabilities in contrast to 1.4.3 or higher. Hence, suggestion is to upgrade.


ganeshpms:
When this will be merged we need to use url-phrse as 1.4.7 has vulnerability

PR #613, opened 2021-03-30
Download amqp.json from "rabbitmq-server" repo

xamgore:
The original repository, [rabbitmq-codegen](https://github.com/rabbitmq/rabbitmq-codegen) has been moved to the main unified RabbitMQ mono repository, [rabbitmq-server](https://github.com/rabbitmq/rabbitmq-server). This PR updates the source URL.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
-RABBITMQ_SRC_VERSION=rabbitmq_v3_2_1
+COMMIT_OR_BRANCH=a334b711ad37493cdccdd141079c4af974a9b054
 JSON=amqp-rabbitmq-0.9.1.json
-RABBITMQ_CODEGEN=https://raw.githubusercontent.com/rabbitmq/rabbitmq-codegen
-AMQP_JSON=$(RABBITMQ_CODEGEN)/$(RABBITMQ_SRC_VERSION)/$(JSON)
+AMQP_JSON_LOCATION=https://raw.githubusercontent.com/rabbitmq/rabbitmq-server/$(COMMIT_OR_BRANCH)/deps/rabbitmq_codegen/$(JSON)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,7 +1,6 @@
-RABBITMQ_SRC_VERSION=rabbitmq_v3_2_1
+RABBITMQ_SRC_VERSION=v3.8.x
```
squaremo, line 1: To make builds more repeatable, I think this should target a specific version rather than a branch.
What might happen? Say a change to the spec is introduced between RabbitMQ 3.8.x versions; one automated build of amqplib would have a codec from before the change, and another -- say from a merge commit -- might have a codec with the change. The result would not be a _catastrophe_, but at least an avoidable cause of confusion.

xamgore: Totally agree. Though, I would prefer to change the build phase instead.

1. The `amqp-rabbitmq.json` specification and `lib/defs.js` should be git added to the repository.
   * Only maintainers usually want to update these files, they do so by calling `make clean; make lib/defs.js`.
 
   * The spec changes then can be tracked and reviewed in pull requests (but nobody will dare)
   * Npm will be able to install the package from a github branch (see #571), like
       ```
       "amqplib": "github:squaremo/amqp.node#pull/584/head"
       ```
      Now, after downloading the source files, the `prepare` npm-script is called and may fail, as not all OSs have `make` installed.
2. Prepublish npm-script or CI cron should check a mismatch of the local and remote `rabbitmq-server/master/.../amqp-rabbitmq.json`.

squaremo: Yeah I'm OK with defs.js being committed, so long as there's an automated check that it's correctly generated.

squaremo: NB I would still like to have an exact version of the spec, so that its provenance is recorded. But that still fits with your plan.

xamgore: Done. An exact commit version or a branch name or a tag can be written there.

PR #615, opened 2021-04-04
Drop support for node.js engines below v10

xamgore:
It's a common practice to drop support for non-LTS Node versions, as it allows to reduce the maintenance burden. See more at:
https://nodejs.org/en/about/releases/

* [aws](https://aws.amazon.com/blogs/compute/node-js-14-x-runtime-now-available-in-aws-lambda/) supports nodejs v14, v12, v10 (will not be available soon)
* [gcd](https://cloud.google.com/appengine/docs/standard/nodejs/runtime#node.js-14) supports nodejs v14, v12, v10, v8 (deprecated)

Also, I see artifacts in the code
> something broken about domains in Node.JS v0.8 and mocha

or

```js
function isFloatingPoint(n) {
    return n >= 0x8000000000000000 ||
        (Math.abs(n) < 0x4000000000000
         && Math.floor(n) !== n);
```

instead of `Number.isSafeInteger(n)` (appeared in v0.12).

My future PRs will be based on this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ JSON=amqp-rabbitmq-0.9.1.json
 RABBITMQ_CODEGEN=https://raw.githubusercontent.com/rabbitmq/rabbitmq-codegen
 AMQP_JSON=$(RABBITMQ_CODEGEN)/$(RABBITMQ_SRC_VERSION)/$(JSON)
 
-NODEJS_VERSIONS='0.8' '0.9' '0.10' '0.11' '0.12' '1.8.4' '2.5' '3.3' '4.9' '5.12' '6.17' '8.17' '9.11' '10.21' '11.15' '12.18' '13.14' '14.5' '15.8'
+NODEJS_VERSIONS='10.21' '11.15' '12.18' '13.14' '14.5' '15.8'
 
 MOCHA=./node_modules/.bin/mocha
 _MOCHA=./node_modules/.bin/_mocha
--- a/README.md
+++ b/README.md
@@ -9,9 +9,7 @@
  * [API reference][gh-pages-apiref]
  * [Examples from RabbitMQ tutorials][tutes]
 
-A library for making AMQP 0-9-1 clients for Node.JS, and an AMQP 0-9-1
-client for Node.JS v0.8-0.12, v4-v15, and the intervening io.js
-releases.
+A library for making AMQP 0-9-1 clients for Node.JS, and an AMQP 0-9-1 client for Node.JS v10+.
 
 This library does not implement [AMQP
 1.0](https://github.com/squaremo/amqp.node/issues/63) or [AMQP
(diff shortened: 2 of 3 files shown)


davesag:
Note Node 16 came out today and the `package.json` specifies `<= 15`

I've tried it out with Node 16 and it works fine so expect just removing the upper cap of the `engines` section of `package.json` would be fine.

KristjanTammekivi:
@cressie176 Hi, sorry to mention directly but I haven't seen anything happening, would like to use node 16 but this is blocking

gajus:
@cressie176 If you add me as a collaborator to git / npm (@gajus), I am happy to help with this / future releases.

gajus:
@squaremo Could we get your help releasing this?

squaremo:
I like being compatible with all versions of NodeJS ever (or close to), but I accept this now causes problems and everyone else moved on long ago. I have two questions before this proceeds:

 1. Are there usage stats for NodeJS versions? (alternatively: what proportion of the users of NodeJS does this change exclude?)
 2. What will people who don't use the right version of NodeJS see when they try to use this package?

KristjanTammekivi:
```
$ npm i amqplib
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'amqplib@0.7.1',
npm WARN EBADENGINE   required: { node: '>=0.8 <=15' },
npm WARN EBADENGINE   current: { node: 'v16.1.0', npm: '7.11.2' }
npm WARN EBADENGINE }
```
This is what I currently see, it's only warnings and installs the package. I could swear that it used to refuse to install it.

As for statistics I don't think anyone really has them, however if you're using newest versions of amqplib I'd say you probably somewhat keep up with node.js releases. I don't really care much about the part of this PR but I would like the upper limit to be removed.

kibertoad:
@squaremo This might provide some insights: https://twitter.com/kibertoad/status/1338886973256331267

kibertoad:
I think the tl;dr version is that everything below 4 can be dropped without a second thought. Dropping 4 and 6 would allow you to significantly modernize your code, though, and considering that likely the largest part of it is hopeless legacy whom noone updates anymore, not sure if it's worth to target that either.

squaremo:
OK, let's bite that bullet. @xamgore can you alter the claim in the README.md here too please, then I'll merge it.

xamgore:
@squaremo nailed it, thank you! 👍 

Also, there is one more [ready-to-merge PR](https://github.com/squaremo/amqp.node/pull/567), if you have free time, could you please check it out?

squaremo (approved):
Thank you @xamgore

squaremo:
I went to make a release for this by tagging it, but forgot (again) that it needs a bit of preparation. Later today!

squaremo:
Released in v0.8.0.

PR #620, opened 2021-04-23
added support for 'U' and 'u' to codec

luukwullink:
I had trouble decoding a number (unsigned 16 bit integer) created by a python client (https://github.com/pika/pika) and consumed by this client. I found out that there is no implementation for 'u'.

Let me know if i am missing something, i am not so familiar with nodejs and doing pullrequests ;).

### DIFF
--- a/lib/codec.js
+++ b/lib/codec.js
@@ -279,6 +279,10 @@ function decodeFields(slice) {
         case 'l':
             val = ints.readInt64BE(slice, offset); offset += 8;
             break;
+        case 'u':
+            val = slice.readUInt16BE(offset); offset += 2;
+            break;
+        case 'U':
         case 's':
             val = slice.readInt16BE(offset); offset += 2;
             break;


voicenter:
Hi ...
there are more types besides just the U that are missing ...
please review [https://github.com/squaremo/amqp.node/pull/646](PR-646) that add all of the missing types,
many thanks for this amazing project 
Shlomi Gutman
Voicenter CTO

kibertoad:
@cressie176 Do you think we should merge this? See https://github.com/amqp-node/amqplib/pull/646 on same topic, except in this case it's clear how incorrect data happened.

cressie176:
> @cressie176 Do you think we should merge this? See #646 on same topic, except in this case it's clear how unsupported data was created.

Sorry, didn't properly understand this before. Yes, I think so - if amqplib clients are going to have to consume stuff published by other libraries with different type keys

cressie176:
'u' is now implemented. Decided not to implement 'U' as it is only for amqp 0.9. 
I don't think adding support for 'U' would have any adverse impact so will consider it if it means being able to consume messages published by 0.9 clients though.

PR #624, opened 2021-05-06
feat: switch to native Promise

alexkvak:


kibertoad:
@alexkvak This needs to be synced with master after old Node versions were dropped

lib/channel_model.js
```diff
@@ -74,10 +98,16 @@ C.open = function() {
 
 C.close = function() {
   var self = this;
-  return Promise.fromCallback(function(cb) {
+
+  return Bluebird.fromCallback(function(cb) {
     return self.closeBecause("Goodbye", defs.constants.REPLY_SUCCESS,
                     cb);
   });
+  // return new Promise(function(resolve, reject) {
```
kibertoad, line 106: Should this be commented out?

lib/channel_model.js
```diff
@@ -74,10 +98,16 @@ C.open = function() {
 
 C.close = function() {
   var self = this;
-  return Promise.fromCallback(function(cb) {
+
+  return Bluebird.fromCallback(function(cb) {
```
kibertoad, line 102: Am I misunderstanding something, or promisify could be used here instead of Bluebird?

alexkvak (commented):
> @alexkvak This needs to be synced with master after old Node versions were dropped

Done

lib/channel_model.js
```diff
@@ -167,9 +197,18 @@ C.consume = function(queue, callback, options) {
   // NB we want the callback to be run synchronously, so that we've
   // registered the consumerTag before any messages can arrive.
   var fields = Args.consume(queue, options);
-  return Promise.fromCallback(function(cb) {
+  return Bluebird.fromCallback(function(cb) {
     self._rpc(defs.BasicConsume, fields, defs.BasicConsumeOk, cb);
   })
+  // return new Promise(function(resolve, reject) {
```
kibertoad, line 203: Should this be commented out?

lib/channel_model.js
alexkvak: After dropping old Node version — it could

lib/channel_model.js
alexkvak: I need some help with this.

If I replace `Bluebird.fromCallback` with commented code it seems that `channel.accept` is called _after_ this promise resolved. So the behavior changes and I have no idea why. I'm sure you have more context with this.

lib/channel_model.js
kibertoad: Only Node 10+ is supported in master, so should be fine to use it now.

lib/callback_model.js
```diff
@@ -5,7 +5,6 @@
 'use strict';
 
 var defs = require('./defs');
-var Promise = require('bluebird');
 var inherits = require('util').inherits;
```
jimmywarting, line 8: Unrelated to this PR but i just notices inherits, NodeJS docs mention discourage use of inherits and that class extends should be used instead.

kibertoad: I'll open PR for this

lib/channel_model.js
alexkvak: Replaced with util.promisify

alexkvak: btw .travis.yml in main still builds library for old Node.js versions

kibertoad: Yes, I think it was just missed in https://github.com/squaremo/amqp.node/commit/267dd4e17c93efc755e40a4645510fdf70b9ebb8. Probably Travis needs to be adjusted to only test for Node 10+ as well.

KristjanTammekivi:
Should also edit the types once this is ready https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/amqplib

channel_api.js
```diff
@@ -1,15 +1,10 @@
-var raw_connect = require('./lib/connect').connect;
+const { promisify } = require('util');
+const raw_connect = promisify(require('./lib/connect').connect);
```
jimmywarting, line 2: promisify is undefined

xamgore:
I don't really recommend changing anything until tests are being migrated to jest and a proper [memory leak detector](https://www.npmjs.com/package/jest-leak-detector) is set up. There is a [PR](https://github.com/squaremo/amqp.node/pull/584) that fixes a memory leak, though either it isn't covered by tests or mocha's detector can't catch the case. Anyway, any change in the codebase may bring undesired behaviour in the future, and it may take time to fix and release it. As I may see in the comments here, Bluebird's semantics differs a bit from Node's promise implementation.

Also, @squaremo is quite a busy man and merges PRs if only 100% sure they are ok.

carlhoerberg:
https://github.com/cloudamqp/amqp-client.js/ is a modern zero dependency AMQP library, implementing native promises, among other improvements.

alexkvak:
> https://github.com/cloudamqp/amqp-client.js/ is a modern zero dependency AMQP library, implementing native promises, among other improvements.

Thanks a lot! I'll take a look and return with feedback 👍

mohd-akram:
Use of native promises has been released in v0.10.0 via #689. This can be closed.

PR #628, opened 2021-05-18
Replace safe-buffer dependency with native buffers

kibertoad:
Considering that only Node 10+ is supported now, and Node 6+ has built-in support for safe buffer initialization, which is a drop-in replacement for safe-buffer, this dependency is no longer necessary.

Had to rebuild entire package-lock, because it was (erroneously, I assume) pointing at http://npm.tescloud.com/, which is not public.


jimmywarting:
Something i tend to do is just using Uint8Array whenever possible, more cross platform compatible and more lightweight.
Many of nodes built in functions are now more acceptable of plain Uint8array - feels like they are using buffer less and less.

To use any of the Buffer utilities I switch to using DataView, TextEncoder/decoder, atob instead that also exist on all platforms. But this would be more undertaking to switch to.

jimmywarting (approved):

squaremo (approved):
Yep, makes sense in full. Thanks @kibertoad

PR #631, opened 2021-05-19
Remove old NodeJS versions from CI matrix

squaremo:
Support for Node prior to v10 has been removed. It would be nice to check against older NodeJS versions, but it can't be required by CI.


kibertoad (approved):

squaremo:
Thanks for the review @kibertoad 💯

PR #644, opened 2021-08-20
Bump url-parse library to fix CVE-2021-3664 vulnerability

abdulgit2021:


cressie176:
Thanks for the PR. amqplib v0.9.0 depends on the latest url-parse

PR #645, opened 2021-08-26
GitHub Actions CI

kibertoad:


jimmywarting:
bump

squaremo:
Any idea why this doesn't run the actions? So far as I can tell, it's not restricted to PRs only in my account.

kibertoad:
@squaremo From my understanding, if there were no GitHub Actions before, it needs to be merged before activates.

kibertoad:
this was already implemented

PR #646, opened 2021-09-07
Type tags

MaxKirillov:
Hey there! Recently we faced TypeError: Unexpected type tag "L"...
So added more types according to # 4.2.1 Formal Protocol Grammar (AMQP Protocol Specification https://www.rabbitmq.com/resources/specs/amqp0-9-1.pdf)

### DIFF
--- a/lib/codec.js
+++ b/lib/codec.js
@@ -241,52 +241,85 @@ function decodeFields(slice) {
         var tag = String.fromCharCode(slice[offset]); offset++;
         switch (tag) {
         case 'b':
+        // short-short-int = OCTET
             val = slice.readInt8(offset); offset++;
             break;
+        case 'B':
+        // short-short-uint = OCTET
+             val = slice.readUInt8(offset); offset++;
+             break;
         case 'S':
+        // long-string = long-uint *OCTET
             len = slice.readUInt32BE(offset); offset += 4;
             val = slice.toString('utf8', offset, offset + len);
             offset += len;
             break;
         case 'I':
+        // long-int = 4*OCTET
             val = slice.readInt32BE(offset); offset += 4;
             break;
+        case 'i':
+        // long-uint = 4*OCTET
+            val = slice.readUInt32BE(offset); offset += 4;
+            break;
         case 'D': // only positive decimals, apparently.
+        // decimal-value = scale long-uint
             var places = slice[offset]; offset++;
             var digits = slice.readUInt32BE(offset); offset += 4;
             val = {'!': 'decimal', value: {places: places, digits: digits}};
             break;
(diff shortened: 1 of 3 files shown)


squaremo:
The types as given in the AMQP 0-9-1 spec are problematic -- they are inconsistent, and don't match what servers and clients actually implement (see https://www.rabbitmq.com/amqp-0-9-1-errata.html#section_3). It's a bit of a mess.

I based the type tags here on what RabbitMQ understands, as a matter of practicality. Here's the RabbitMQ server code:  https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_binary_parser.erl#L30; and here's the Java client (the longest running client, iirc): https://github.com/rabbitmq/rabbitmq-java-client/blob/main/src/main/java/com/rabbitmq/client/impl/ValueReader.java#L158. Note the server code understands 'L', but treats it the same as (lowercase L) `l` and reads a signed integer.

I'm interested to know how you got a field table with an unexpected type tag -- do you know which client implemented encoded it?

cressie176:
Tidy up

ebrinette:
Hello !

To answer your question, this is used by the [rabbitmq-c](https://github.com/alanxz/rabbitmq-c) library.

You can add AMQP_FIELD_KIND_U16 in the header table.

What's currently blocking #620 ?

lib/codec.js
```diff
@@ -241,53 +241,86 @@ function decodeFields(slice) {
...
+        // double = 8*OCTET; rfc1832 XDR double
             val = slice.readDoubleBE(offset); offset += 8;
             break;
         case 'f':
+        // float = 4*OCTET; IEEE-754
             val = slice.readFloatBE(offset); offset += 4;
             break;
         case 'l':
+        // long-long-uint = 8*OCTET
+            val = ints.readUInt64BE(slice, offset); offset += 8;
```
cressie176, line 298: According to https://www.rabbitmq.com/amqp-0-9-1-errata#section_3 'l' means a signed 64-bit integer for RabbitMQ, not an Unsigned 64-bit integer.

isaacharrisholt:
I got an unexpected type tag `i` when using `lapin` in Rust, sending a `LongUInt` header. Would be great to see this in :)

cressie176:
Two of the type tags 'L' and 'U' in this PR are for amqp 0.9, and not present in amqp 0.9.1 
The 'l' type tag was also for amqp 0.9 but has a different meaning in amqp 0.9.1
The others have since been implemented in other pull requests.
As such I'm going to close without merging, but if the absence of 'L' and 'U' causes an issue I will consider adding them.

PR #647, opened 2021-09-18
Close connection to server on connect errors

luddd3:
A minimal pull request to fix open socket leak after e.g. authentication error on connect. This is a part of the fix in #584, but I have only kept the most essential part and added a test.

By keeping this as small as possible, I hope to remove any hesitance so that it can be merged.

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -152,8 +152,12 @@ function connect(url, socketOptions, openCallback) {
       if (timeout) sock.setTimeout(0);
       if (err === null) {
         openCallback(null, c);
+      } else {
+        // The connection isn't closed by the server on e.g. wrong password
+        sock.end();
+        sock.destroy();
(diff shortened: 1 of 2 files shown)


Uzlopak:
Never heard of the concept of bike shedding? :D

luddd3:
@kibertoad Sorry for pinging, but I saw in another thread that you do some work in this repo. Can you please review this PR too?

kibertoad (approved):

kibertoad:
Thank you!

kibertoad:
@luddd3 CI is failing on main now, could you please take a look?

luddd3:
@kibertoad Fixed in #667
I think there are many steps during setup that the `openCallback` might be called more than once. For example, what if the socket is abruptly closed in-between the steps? Then `bail` would be called multiple times.

Issue #657, opened 2021-11-04
Mux.js Usage of .concat questionable


Uzlopak:
I think we could improve the performance and memory footprint of Mux.js if we  change
https://github.com/squaremo/amqp.node/blob/243aef8b6da766a3b9908bc50b2be4859eef43a6/lib/mux.js#L82

to this:
```javascript
Array.prototype.push.apply(this.oldStreams, this.newStreams);
```

or 
```javascript
this.oldStreams.push(...this.newStreams);
```

.concat is creating a new Array. But if we directly push the elements of newStreams-Array directly into the oldStreams-Array, we would actually reduce the memory overhead and probably would gain some minor performance gains.

carlhoerberg:
If you're looking for performance i would recommend https://github.com/cloudamqp/amqp-client.js/. We've optimized it quite heavily and are open to more improvements.

Uzlopak:
@carlhoerberg 

Do you have benchmarks for your claim :)?

carlhoerberg:
Yes, readme updated

Uzlopak:
Very interesting. Thank you.

Uzlopak:
@carlhoerberg 

Can you supply the benchmarks also into the repo? Maybe something, where you have to spawn  a rabbitmq-server via docker-compose?

cressie176:
Switching to from concat to Array.prototype.push.apply makes a substantial difference to performance...

```javascript
const assert = require('assert');

class Appender {
  constructor(size) {
    this.size = size;
    this.a = [];
    this.b = new Array(size).fill().map(Math.random);
  }
}

class Push extends Appender {
  test() {
    Array.prototype.push.apply(this.a, this.b);
  }
}

class Concat extends Appender {
  test() {
    this.a = this.a.concat(this.b);
  }
}

function test(variant, iterations) {
  const before = Date.now();
  for (let i = 0; i < iterations; i++) {
    variant.test();
  }
  const after = Date.now();
  assert(variant.a.length, variant.size * iterations)
  return after - before;
}

function testMany(variant, iterations) {
  return new Array(iterations)
    .fill()
    .map(() => test(variant, iterations))
    .reduce((total, duration) => (total + duration), 0);
}

const size = 10;
const iterations = 100;

const push = testMany(new Push(size), iterations);
const concat = testMany(new Concat(size), iterations);
console.log({ push, concat })
```

```
{ push: 4, concat: 1964 }
```
[...]

PR #658, opened 2021-11-04
use .push.apply instead of concat in Mux _readIncoming

Uzlopak:
Also
remove unused require
remove unused self variable

#657

Should reduce memory footprint and performance

### DIFF
--- a/lib/mux.js
+++ b/lib/mux.js
@@ -77,7 +77,7 @@ Mux.prototype._readIncoming = function() {
   }
   else { // ran out of room
     assert(this.newStreams.length > 0, "Expect some new streams to remain");
-    this.oldStreams.push.apply(this.newStreams);
+    Array.prototype.push.apply(this.oldStreams, this.newStreams);
     this.newStreams = [];
   }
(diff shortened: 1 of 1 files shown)


lib/mux.js
```diff
@@ -79,7 +77,7 @@ Mux.prototype._readIncoming = function() {
   }
   else { // ran out of room
     assert(this.newStreams.length > 0, "Expect some new streams to remain");
-    this.oldStreams = this.oldStreams.concat(this.newStreams);
+    Array.prototype.push.apply(this.oldStreams, this.newStreams);
```
kibertoad, line 80: Is there a reason why prototype push is used instead of `this.oldstreams.push`?

Uzlopak: Well I can not test it, but you could do a this.oldstreams.push.apply(this.newstreams)?

I think I did this to make it clear that it does not instantiate a intermediary Array

Uzlopak: https://stackoverflow.com/questions/15444242/why-doesnt-array-push-apply-work

I think this was the reason. I don't want to reference this.oldstreams twice.

PR #659, opened 2021-11-04
call unregisterConsumer in handleCancel

### LINKED ISSUE #585: Memory leak by binding then remove Queue
Hello
If you create a queus, then create a binding and then delete a queu, you will see that the application memory grows, but if you unbind before deleting the queue, then the problem goes away, is this standard behavior?
There is an assumption that the bindings in the libraries are saved even after the queue is deleted, it seems to me this is a bug, because the binding to the queue should not be stored without this queue

Uzlopak:
This PR adds a call to unregisterConsumer in handleCancel. This ensures, that the consumer gets removed from this.consumers.

Fixes #585

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -493,5 +493,7 @@ BaseChannel.prototype.handleDelivery = function(message) {
 };
 
 BaseChannel.prototype.handleCancel = function(fields) {
-  return this.dispatchMessage(fields, null);
+  var result = this.dispatchMessage(fields, null);
+  this.unregisterConsumer(fields.consumerTag);
+  return result;
 };


kibertoad:
Can you also add a test for this?

Uzlopak:
@kibertoad 
I have to see if I can write a unit test. I think when I implemented this I thought that @squaremo would check if the code makes change.

Do you have the permissions to merge?
Alot of stuff we could be improved.

Have a look at #658 . It think this could actually be solve some memory issues as it should potentially not create intermediate arrays.

kibertoad:
@Uzlopak I do!

Issue #656, opened 2021-11-04
Use Maps or WeakMaps instead of Objects for better V8 performance


Uzlopak:
I think we should implement LookupObject e.g. BaseChannel.consumers as Maps or WeakMaps instead of Objects. V8 creates for each shape of an Object a new V8-Class of that Object. So if you dynamically consume and cancel queues V8 should become slower and slower. If we use Maps it should have not that effect.

carlhoerberg:
We do this in https://github.com/cloudamqp/amqp-client.js/ already

cressie176:
published in amqplib@0.9.1 Thank you @Uzlopak

PR #660, opened 2021-11-04
Use Map instead of Object for BaseChannel.consumers

Uzlopak:
This PR uses a Map instead of an Object for BaseChannel.consumers. It should improve the performance while running in node, as the amount of V8-HiddenClasses will be reduced. 

Use Case: If in your implementation alot of consumers get created and canceled, alot of unique V8-HiddenClasses will be created in V8 at runtime. 

Information regarding HiddenClasses:
https://v8.dev/blog/slack-tracking

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -460,7 +460,7 @@ C.onBufferDrain = function() {
 // low-level machinery.
 function BaseChannel(connection) {
   Channel.call(this, connection);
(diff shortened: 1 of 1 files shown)


Uzlopak:
Should fix #656

cressie176:
Thank you @Uzlopak. I've merged into main and will close when I next publish a release.

PR #661, opened 2021-11-05
transform throws string into throw Error

Uzlopak:
@squaremo 
While analyzing amqplib in sonarqube I found that some throws do tot throw an Error-Object but merely a string. 

If you now run the tests, two of them fail. 

```
1) Consume from non-queue invokes error k
2) bad delivery
```

Responsible for failing these tests is the change in line 346. 

I suspect, that this is a valid code smell but I quite dont understand how I could fix this.

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -268,8 +268,8 @@ function acceptDeliveryOrReturn(f) {
   var event;
   if (f.id === defs.BasicDeliver) event = 'delivery';
   else if (f.id === defs.BasicReturn) event = 'return';
-  else throw fmt("Expected BasicDeliver or BasicReturn; got %s",
-                 inspect(f));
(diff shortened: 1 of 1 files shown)


lib/channel.js
```diff
@@ -343,7 +343,7 @@ function acceptMessage(continuation) {
         return content;
       }
     }
-    else throw "Expected content frame after headers"
+    else throw new Error("Expected content frame after headers");
```
Uzlopak, line 346: This line results in failing unit tests as described in the initial post

cressie176:
I backed this one out. The channel's [acceptMessageFrame function](https://github.com/amqp-node/amqplib/blob/main/lib/channel.js#L249-L251) uses the error type to differentiate between unexpected frames and other types of errors.

PR #667, opened 2021-12-28
fix: done called twice on invalid options

luddd3:
Fixes CI with node version 10.

The problem was that the steps in `afterStartOk` weren't aborted even though an error was detected and `bail` was called. Thus the callback was called at least twice.


kibertoad (approved):

kibertoad:
appreciate the quick fix :)

luddd3:
Sorry for the inconvenience and thank you for reviewing/merging =)

PR #668, opened 2022-01-30
Add example for RabbitMQ stream queues

Gsantomaggio:
It shows how to use the stream queue with the AMQP client.

See: 
- https://www.rabbitmq.com/stream.html
- https://blog.rabbitmq.com/posts/2021/07/rabbitmq-streams-overview

### DIFF
--- a/examples/stream_queues/README.md
+++ b/examples/stream_queues/README.md
@@ -1,0 +1,15 @@
+RabbitMQ Stream Examples 
+---
+The [stream queues](https://www.rabbitmq.com/streams.html) are available starting from RabbitMQ 3.9
+
+These examples show how to use stream queues with the lib.
+
+Send a message to a stream queue
+```
+./send_stream.js   
+```
+
+Reveice all the messages from stream queue:
+```
+./receive_stream.js 
+```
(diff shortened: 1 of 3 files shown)


carlhoerberg:
The stream plugin is not required for this example, only Rabbitmq 3.9 which has stream queue support built-in, without the need for a plugin. The plugin only adds custom protocol support.

Gsantomaggio:
Ops sorry you are right! I changed the description! 
thank you

examples/stream_queues/receive_stream.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+
+var amqp = require('amqplib');
```
kibertoad, line 3: can you please rewrite to use modern style with `const`/`let` and `async/await`?

examples/stream_queues/README.md
```diff
@@ -0,0 +1,15 @@
...
+The [stream queues](https://www.rabbitmq.com/streams.html) are available starting from RabbitMQ 3.9
+
+These examples show how to use stream queues with the lib.
+
+Send a message to a stream queue
+```
+./send_stream.js   
+```
+
+Reveice all the messages from stream queue:
```
kibertoad, line 12:
```suggestion
Receive all the messages from stream queue:
```

kibertoad (approved):

PR #678, opened 2022-04-09
Replace url-parse lib with native URL interface

jonaswalden:
Engine specified to node >= 10.
Should be no need for a lib to polyfill the `URL` interface or implement the legacy `url.parse` in a browser environment.


lib/connect.js
```diff
@@ -104,37 +103,30 @@ function connect(url, socketOptions, openCallback) {
...
 
-    var config = {
-      locale: url.locale,
-      channelMax: url.channelMax,
-      frameMax: url.frameMax,
-      heartbeat: url.heartbeat,
-    };
-
-    fields = openFrames(url.vhost, config, sockopts.credentials || credentials.plain(user, pass), extraClientProperties);
+    fields = openFrames(url.vhost, url.searchParams, sockopts.credentials || credentials.plain(user, pass), extraClientProperties);
```
jonaswalden, line 121: CAUTION (maybe) `vhost` used here is not a part of a `URL` instance. 
Then again it wasn't a part of the result of `url-parse` either so this was `undefined` anyway. The only place I found where it wasn't `undefined` was in the [bad URL test](https://github.com/amqp-node/amqplib/blob/main/test/connect.js#L66).

cressie176:
Thanks @jonaswalden.

https://github.com/amqp-node/amqplib/commit/7acbfd72330b835e1ffdad3daf78c222d1256142 details why url-parse was introduced in the first place. I suspect that now amqplib only supports node 10 and up, the standard url parser will be fine. Unless this isn't so, I'm in favour of this PR - the fewer 3rd party libs we import the better

kibertoad:
@jonaswalden Could you please rebase this?

jonaswalden:
@cressie176 wow, ok. good to know! seems like that's covered https://github.com/BonnierNews/amqplib/blob/main/test/connect.js#L54

jonaswalden:
hold on, I screwed something up

lib/connect.js
```diff
@@ -104,37 +103,30 @@ function connect(url, socketOptions, openCallback) {
 
   var protocol, fields;
   if (typeof url === 'object') {
-    protocol = (url.protocol || 'amqp') + ':';
+    protocol = url.protocol || 'amqp:';
```
kibertoad, line 106: `protocol` no longer gets `:` appended to it, is it intentional?

lib/connect.js
kibertoad: this sounds like a bug. can we fix it? `vhost` is important

test/connect.js
```diff
@@ -29,27 +29,27 @@ suite("Credentials", function() {
   }
 
   test("no creds", function(done) {
-    var parts = urlparse('amqp://localhost');
+    var parts = new URL('amqp://localhost');
     var creds = credentialsFromUrl(parts);
     checkCreds(creds, 'guest', 'guest', done);
   });
   test("usual user:pass", function(done) {
```
kibertoad, line 36: can we add test case for the vhost, using the 
`amqp://${config.username}:${config.password}@${config.hostname}:${config.port}/${config.vhost}` structure?

kibertoad:
@jonaswalden Do you need any help with finishing this?

PR #680, opened 2022-04-18
Simplify publishing releases to npm

cressie176:
I've added a GitHub action for automatically publishing releases to npm. The envisioned workflow is as follows...

1. Work normally, merging PRs, until ready to release etc
2. Check that the main branch build is green
3. Check that your local environment is clean and on main and up to date
4. Use `npm version major|minor|patch` to update package.json and create a tag, e.g. v0.8.1
5. Use `git push && git push --tags` to push both the package.json change and the new tag
6. Create a new release from the release in the GitHub UI
7. Wait for the new Publish action to publish the release to npm

For this to work we will need a repository secret configured with an NPM access token. It will be better to create this from a new npm robot account, as there's no way to restrict the access token to just the amqplib module.

One concern I have is that a GitHub action could be modified to expose the secret, which could then be used to publish a vulnerable version of amqplib. Since secrets are not exposed to forks, a maintainer would have to merge the malicious action first though. We should only accept this PR if we are happy with this risk.

### DIFF
--- a/lib/mux.js
+++ b/lib/mux.js
@@ -13,117 +13,114 @@ var assert = require('assert');
 var schedule = (typeof setImmediate === 'function') ?
   setImmediate : process.nextTick;
 
-function Mux(downstream) {
-  this.newStreams = [];
-  this.oldStreams = [];
-  this.blocked = false;
-  this.scheduledRead = false;
+class Mux {
+  constructor (downstream) {
+    this.newStreams = [];
+    this.oldStreams = [];
+    this.blocked = false;
+    this.scheduledRead = false;
 
-  this.out = downstream;
-  var self = this;
-  downstream.on('drain', function() {
-    self.blocked = false;
-    self._readIncoming();
-  });
-}
+    this.out = downstream;
+    var self = this;
+    downstream.on('drain', function () {
+      self.blocked = false;
+      self._readIncoming();
+    });
+  }
 
-// There are 2 states we can be in:
+  // There are 2 states we can be in:
+  // - waiting for outbound capacity, which will be signalled by a
+  // - 'drain' event on the downstream; or,
+  // - no packets to send, waiting for an inbound buffer to have
+  //   packets, which will be signalled by a 'readable' event
+  // If we write all packets available whenever there is outbound
+  // capacity, we will either run out of outbound capacity (`#write`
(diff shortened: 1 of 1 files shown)


.github/workflows/node-ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       rabbitmq:
-        image: rabbitmq
+        image: rabbitmq:3.9-alpine
```
cressie176, line 17: alpine should be faster to download. I also noticed some odd problems with the ubuntu version where it looked like ubuntu wasn't able to install rabbitmq.

.github/workflows/node-publish.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: https://registry.npmjs.org/
+
+      - run: make lib/defs.js
+
+      - run: npm publish --dry-run
```
cressie176, line 22: Will remove --dry-run after testing

.github/workflows/node-ci.yml
kibertoad: Any reason why you are not using the latest version?

cressie176: Good spot, I'll change to 3-alpine.

kibertoad, line 17:
```suggestion
        image: rabbitmq:3-alpine
```

.github/workflows/node-publish.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
```
kibertoad, line 17:
```suggestion
          node-version: '18.x'
```

kibertoad:
@cressie176 Do you need any help with this?

cressie176:
Hi @kibertoad,

Unfortunately I don't have sufficient access to the GitHub repo to add a secret.

PR #689, opened 2022-05-19
Use native promises

mohd-akram:


cressie176:
Thanks @mohd-akram, I'll review shortly

lib/channel_model.js
```diff
@@ -286,6 +282,7 @@ class ConfirmChannel extends Channel {
             if (err === null) resolve();
             else reject(err);
           };
+          if (!this.pending) unconfirmed[index](new Error('channel closed'))
```
kibertoad, line 285: Where is this coming from?

test/channel.js
```diff
@@ -75,13 +75,11 @@ var DELIVER_FIELDS = {
...
 };
 
-function open(ch) {
-  return Promise.try(function() {
-    ch.allocate();
-    return Promise.fromCallback(function(cb) {
-      ch._rpc(defs.ChannelOpen, {outOfBand: ''}, defs.ChannelOpenOk, cb);
-    });
-  });
+async function open(ch) {
```
kibertoad, line 78: why make it async?

lib/channel_model.js
mohd-akram: This does the same as [this](https://github.com/amqp-node/amqplib/blob/aa46d7681a5811a8dcf99ee9a711f03d0a1e7f29/lib/channel.js#L38) and was necessary for a [test](https://github.com/amqp-node/amqplib/blob/aa46d7681a5811a8dcf99ee9a711f03d0a1e7f29/test/channel_api.js#L583). 'close' is emitted before `waitForConfirms` is called so it wasn't being handled there. `this.pending` is set to null when the channel is closed (in `Channel._rejectPending`) so that's used to check.

test/channel.js
mohd-akram: My understanding of `Promise.try` is it makes all exceptions, synchronous or asynchronous, to get thrown in `.catch`. Making the function async does the same thing by wrapping the whole code in a promise.

cressie176: I'm confused by how the open was originally being used. There are some cases when `then` is never called, so the ChannelOpen rpc may execute after the code has continued.

I'm also not sure why Promise.try was required in the first place. ch.allocate can throw a synchronous error, but never does when I run the test suite. If we're uncomfortable with making the function async, then both the following appear to work...

```js
// Decide that Promise.try wasn't necessary to start with
function open(ch) {
  ch.allocate();
  return promisify(function(cb) {
    ch._rpc(defs.ChannelOpen, {outOfBand: ''}, defs.ChannelOpenOk, cb);
  })();
}
```

```js
// Emulate Promise.try without async
function open(ch) {
  return new Promise((resolve, reject) => {
    try {
      ch.allocate();
      ch._rpc(defs.ChannelOpen, {outOfBand: ''}, defs.ChannelOpenOk, (err, channel) => {
        if (err) return reject(err);
        resolve(channel);
      });
    } catch (err) {
      reject(err);
    }
  })
}
```

kibertoad: async is redundant if we are returning Promise already, which I believe we are

lib/channel_model.js
kibertoad: this sounds like it could result in memory leaks. when do we dispose of `unconfirmed`?

test/channel.js
cressie176: In the original code if ch.allocate throws an error while still synchronous, it will be thrown rather than rejected. Promise.try prevents this by wrapping the supplied function in a try/catch and rejecting anyway. That's the only reason for using Promise.try that I can see, but it doesn't seem to be necessary in this case

kibertoad:
Could you also fix conflicts?

lib/channel_model.js
mohd-akram: True, I guess. Ostensibly it's disposed in the 'close' handler in `lib/channel.js`. However, due to the check in the while loop it exits midway (`unconfirmed` seems to have values of `false`). I've updated the code to match what's in the 'close' handler.

test/channel.js
mohd-akram: I've removed the `async` since it seems unnecessary.

mohd-akram:
Updated.

cressie176:
all good @kibertoad?

kibertoad (approved):

cressie176:
Published in amqplib@0.10.0

Thank you @mohd-akram

PR #721, opened 2023-06-08
Convert CommonJS to ESM

revosw:
This commit aims to convert the entire project to ESM from CommonJS. The ESM journey started with specifying `{ "type": "module" }` inside `package.json` and fixing errors as I went.

Notable changes:
- Because there are no synchronous equivalents to `require()` for ESM, the dependency is statically imported at the top, even though there are no guarantees that it will be used.
- `require('stream').Duplex || require('readable-stream/duplex')` is a pattern used a couple places. This new change does not try to import from `readable-stream/duplex` at all, only from `stream`

There are currently 12 failing tests.

Technical changes relating to ESM conversion:
- convert `require()` to `import from`
- remove `'use strict';`
- convert `module.exports` to `export`
- convert all `var` to `const`/`let`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,8 @@ clean:
 lib/defs.js: $(UGLIFY) bin/generate-defs.js bin/amqp-rabbitmq-0.9.1.json
 	(cd bin; node ./generate-defs.js > ../lib/defs.js)
 	$(UGLIFY) ./lib/defs.js -o ./lib/defs.js \
-		-c 'sequences=false' --comments \
-		-b 'indent-level=2' 2>&1 | (grep -v 'WARN' || true)
+		--comments \
+		-b 'indent_level=2' 2>&1 | (grep -v 'WARN' || true)
 
 test: lib/defs.js
 	$(MOCHA) --check-leaks -u tdd --exit test/
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Still working on:
 ## Callback API example
 
 ```javascript
-const amqplib = require('amqplib/callback_api');
+import * as amqplib from 'amqplib/callback_api';
 const queue = 'tasks';
 
 amqplib.connect('amqp://localhost', (err, conn) => {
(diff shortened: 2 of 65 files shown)


revosw:
This is a big PR, and I totally understand if conversion to ESM is not viable right now. This can instead be taken as inspiration for some later date, and as a testament to how smoothly the conversion was (I began the migration no more than 7 hours ago.)

I am not sure if there are any public metrics on which node version the library consumers are running. When the majority of consumers are running Node.js v14+ (released May 2020), and all 12 tests are passing, this PR can proceed.

revosw:
To give an idea on which tests are failing:
```
1) sending messages
   send to queue and consume noAck:
   Uncaught ReferenceError: headers_encoded is not defined
   at encodeBasicProperties (file:///mnt/c/dev/js/amqplib/lib/defs.js:5488:15)

2) sending messages
   send to queue and consume ack:
   Uncaught ReferenceError: headers_encoded is not defined
   at encodeBasicProperties (file:///mnt/c/dev/js/amqplib/lib/defs.js:5488:15)

...4 more tests with the same "not defined" error...

6) heartbeats
   detect lack of heartbeats:
   Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/mnt/c/dev/js/amqplib/test/channel.js)

...3 more tests with the same "Timeout of 2000ms" error...

11) Roundtrip properties
    <BasicProperties> roundtrip:
    » Threw ReferenceError: contentType_len is not defined
    at encodeBasicProperties (file:///mnt/c/dev/js/amqplib/lib/defs.js:5472:22)

12) Content framing
    Adhere to frame max:
    » Threw ReferenceError: contentType_len is not defined
    at encodeBasicProperties (file:///mnt/c/dev/js/amqplib/lib/defs.js:5472:22)
```

revosw:
Combining efforts with #712 would complete the es6 migration

cressie176:
As mentioned in https://github.com/amqp-node/amqplib/issues/720, I'm still not sure what practical benefits migrating to ESM will have. On the flip side it would make amqplib unusable for older versions of node.

revosw:
The benefits for this library specifically are merely for the ecosystem to further converge on a single module system. I agree with you that migration is not currently viable, as there are an undocumented amount of people running ESM-incompatible versions of node.

PR #724, opened 2023-06-13
Configure assertExchange passive through options

MPuust:


cressie176:
Hi @MPuust,

Adding the passive option to `channel.assertQueue` or `channel.assertExchange` means that when passive is true, any other options are meaningless, which I suspect the reason that passive is hard coded and the additional `channel.checkQueue` and `channel.checkExchange` methods exist.

Any reason for not using them?

Also you have reused the `durable` option instead of creating a new `passive` one, which I suspect wasn't your intention.

PR #725, opened 2023-06-16
Connect using instantiated socket.

HiImGiovi:
Connect method modified to accept an already instantiated socket as input.

### DIFF
--- a/examples/stream_queues/README.md
+++ b/examples/stream_queues/README.md
@@ -27,14 +27,14 @@ channel.consume(queue, onMessage, {
 
 RabbitMQ supports six different types of offset, however specifying them can be
 
-| Offset Type | Example Value                                                  | Notes |
-|-----------|------------------------------------------------------------------|-------|
-| First     | `{ 'x-stream-offset': 'first' }`                                 | Start from the first message in the log |
-| Last      | `{ 'x-stream-offset': 'last' }`                                  | Start from the last "chunk" of messages (could be multiple messages) |
-| Next      | `{ 'x-stream-offset': 'next' }`                                  | Start from the next message (the default) |
-| Offset    | `{ 'x-stream-offset': 5 }`                                       | a numerical value specifying an exact offset to attach to the log at |
-| Timestamp | `{ 'x-stream-offset': { '!': 'timestamp', value: 1686519750 } }` | a timestamp value specifying the point in time to attach to the log at. The timestamp must be the number of seconds since 00:00:00 UTC, 1970-01-01. Consumers can receive messages published a bit before the specified timestamp. |
-| Interval  | `{ 'x-stream-offset': '1h' }`                                    | the time interval relative to current time to attach the log at. Valid units are Y, M, D, h, m and s |
+| Offset Type | Example Value                                          | Notes |
+|-----------|----------------------------------------------------------|-------|
+| First     | `{'x-stream-offset':'first'}`                            | Start from the first message in the log |
+| Last      | `{'x-stream-offset':'last'}`                             | Start from the last "chunk" of messages (could be multiple messages) |
+| Next      | `{'x-stream-offset':'next'}`                             | Start from the next message (the default) |
+| Offset    | `{'x-stream-offset':5}`                                  | a numerical value specifying an exact offset to attach to the log at |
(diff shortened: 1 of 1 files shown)


cressie176:
Hi @HiImGiovi,
Thank you for taking the time to submit a PR. What's the reason you need to supply your own socket please?

HiImGiovi:
Hi Stephen,
it's nice to hear from you.
I had to do this for my application since I had to connect to RabbitMQ on-premise using socks protocol.
When connecting to the socks proxy I am receiving a raw socket and this is why I am passing a raw socket which has the information for reaching the host of the RabbitMQ server via a proxy server.

HiImGiovi:
Sorry for the bad form of the PR since it's my first time contributing to open source and I am not a JS developer :) so thank you for the advices.
Regarding point 1 and 2 I will try to write a new method if you'll find this approach is correct from your research.
About point 3, I had to do that to install the package using "npm install git+<git repo url>". Also can you give me an hint on how do I build the project on windows if possible? If I clone the repo and run "make" or "npm i" i receive error on lib/defs.js not found.
About 4 you're probably right.

cressie176:
Hi @HiImGiovi, no need to apologise - I appreciate the work you're putting in to make amqplib better.

Regarding Open Source contribution etiquette, I find opening an issue with an explanation of the problem you're facing, and if possible, outlining a proposed solution is the best first step. That way a discussion can be had in advance of any work done. 

When making a PR, I try to keep it just to the things that are relevant to the issue. The more minimal/targeted the changes are, the more likely the PR will be merged quickly.

> I had to do that to install the package with npm from the git repo

This makes sense, thank you for explaining

> Also can you give me an hint on how do I build the project on windows if possible? If I clone the repo and run "make" or "npm i" i receive error on lib/defs.js not found

I'm afraid I don't have a windows machine, so can't be much help. On my mac I can just run `make`, which executes the first target ("all") in the makefile by default. This downloads the RabbitMQ/amqp 0.9.1 definitions and generates the JavaScript equivalents. I don't understand why you receive a missing file error since the "all" target does not check for this files existence. The "clean" target attempts to remove it without a `force` parameter, which would cause this error. Is there any way "clean" is running instead of "all"?

cressie176:
Any reason why you closed this @HiImGiovi?

HiImGiovi:
Hi Stephen,
I didn't mean to, I wanted to clean the history by creating just a commit
with the new method to create the client from an existing socket removing
the things not relevant to the issue as you suggested.
I didn't know that making an hard reset followed by a push would have
closed the PR.
Anyway I'm going to push a new commit soon.

PS: also I can confirm that cloning the repo and running an "npm i" from
Windows does not work. I tried with WSL and it works fine from Ubuntu.
I probably think that the makefile needs to be adapted for each OS.

Simone

cressie176:
Hi Simone,

OK, no problems. I've been thinking about what approach to take adding the socket, and this is where I've landed...

I agree that providing the socket is the best solution, however better done through an alternative method other than connect, since when providing the socket, (a) the connection has already been established and (b) some of the configuration such as the protocol, hostname, port and socket options would be redundant.

Therefore I suggest adding an additional callback_api function called "establish", which takes three parameters

1. the raw socket
2. options for things like vhost, credentials, heartbeat, frameMax, channelMax and nested socketOptions, etc
3. a callback. 

and uses the raw socket when completing the handshake with the RabbitMQ broker. The existing [onConnect](https://github.com/amqp-node/amqplib/blob/main/lib/connect.js#L143-L162) callback should be extracted and refactored to accept a socket and other variables referenced from the surrounding scope.

Tests should validate a connection can be established using an existing socket, and that the supplied options are honoured. The [channel_api](https://github.com/amqp-node/amqplib/blob/main/channel_api.js) will also need to be updated to expose the new establish method through the promise based interface. Finally the channel_api documentation will also need to be updated.

I'm not sure how easy the above will be, but if you're still happy to get involved, I look forward to the PR

cdromas:
Hello @cressie176, @HiImGiovi, any news about this subject ?
We need to connect to a RabbitMQ instance passing thought a corporate http proxy and this way could be interesting for us.

cressie176:
Hi @cdromas 
I'm not sure how you could tunnel amqp through an http proxy. Some discussion [here](https://github.com/amqp-node/amqplib/issues/394). I might be missing something though.

PR #728, opened 2023-06-25
Add ESLint

kibertoad:


kibertoad:
@cressie176 Could you please take a look?

.eslintrc.json
```diff
@@ -0,0 +1,28 @@
...
+  "rules": {
+    "no-unused-vars": [
+      "warn",
+      {
+        "argsIgnorePattern": "^_",
+        "varsIgnorePattern": "^_",
+        "caughtErrorsIgnorePattern": "^_"
+      }
+    ],
+    "no-cond-assign": "warn",
```
cressie176, line 25: I suggest we change this to "off" given it's used fairly extensively. Warnings are just annoying.

lib/channel_model.js
```diff
@@ -179,6 +179,7 @@ class Channel extends BaseChannel {
       this.unregisterConsumer(consumerTag);
       return ok.fields;
     });
+    return ok
```
cressie176, line 182: Returning ok looks like a change in behaviour. Was this previously an omission or just made to appease a linting rule? 
What does the Basic Cancel RPC return?
Were the api docs updated?

lib/codec.js
```diff
@@ -105,7 +105,7 @@ function encodeFieldValue(buffer, value, offset) {
     var start = offset;
     var type = typeof value, val = value;
     // A trapdoor for specifying a type, e.g., timestamp
-    if (value && type === 'object' && value.hasOwnProperty('!')) {
+    if (value && type === 'object' && Object.prototype.hasOwnProperty.call(value, '!')) {
```
cressie176, line 108: Sometimes I hate javascript!

package.json
```diff
@@ -19,12 +19,14 @@
...
   "devDependencies": {
     "claire": "0.4.1",
+    "eslint": "^7.32.0",
     "mocha": "^9.2.2",
     "nyc": "^15.1.0",
     "uglify-js": "2.8.x"
   },
   "scripts": {
     "test": "make test",
+    "lint": "eslint \"lib/**/*.js\" channel_api.js callback_api.js",
```
cressie176, line 29: Does this catch the generated files? Maybe they should also be ignored (although if they pass, no harm being done currently).

cressie176 (commented):
In addition to the comments, what do you think about a future PR which uses https://www.npmjs.com/package/lint-staged to only check staged files, and potentially running automatically on a commit hook? I'm conscious the more dependencies we add, the more maintenance we incur though.

PR #734, opened 2023-07-14
Add support for x-queue-type option

Gasol:
This commit adds support for the `queueType` option in the `Channel#assertQueue` function. The `queueType` option allows users to specify the type of the queue. Just likes the other options of RabbitMQ extensions can also be supplied as options.

### DIFF
--- a/lib/api_args.js
+++ b/lib/api_args.js
@@ -60,6 +60,7 @@ Args.assertQueue = function(queue, options) {
   setIfDefined(argt, 'x-max-priority', options.maxPriority);
   setIfDefined(argt, 'x-overflow', options.overflow);
   setIfDefined(argt, 'x-queue-mode', options.queueMode);
+  setIfDefined(argt, 'x-queue-type', options.queueType);
 
   return {
     queue: queue,


cressie176:
Hi @Gasol,
Are you aware it is possible to specify x-queue-type via the arguments sub doc, e.g. 

`channel.assertQueue('q', { arguments: { 'x-queue-type': 'classic' } }`

Gasol:
> `channel.assertQueue('q', { arguments: { 'x-queue-type': 'classic' } }`

Yes, I am currently using the same arguments you mentioned. Initially, I attempted to use camel case for the property name as `queueType`, but it didn't work as expected. I discovered that the documentation mentioned the possibility of specifying other extension arguments without the `x-` prefix.

https://amqp-node.github.io/amqplib/channel_api.html#channel_assertQueue
> RabbitMQ extensions can also be supplied as options. These typically require non-standard x-* keys and values, sent in the arguments table; e.g., 'x-expires'. When supplied in options, the x- prefix for the key is removed; e.g., 'expires'. Values supplied in options will overwrite any analogous field you put in options.arguments.


However, after some trial and error, I take a look into the source code and identified the root cause of the issue. This made me contemplate the possibility that others might end up wasting their time trying to resolve the same problem again just like me. As a result, The PR was born.

cressie176:
Thanks for the info @Gasol, makes sense to me

cressie176:
I've decided not to merge, but will review the documentation around queue arguments

PR #744, opened 2023-11-06
Connection: forcibly close stream if close() is called while the connection is blocked

SimonWoolf:
Because of how rabbit's connection.blocked mode works, if the connection is blocked, the client will not receive a close-ok unless/until the connection becomes unblocked, which may be some time away, or never. Current behaviour just leaves the connection open indefinitely. This instead closes it immediately.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -50,6 +50,7 @@ class Connection extends EventEmitter {
     this.recvSinceLastCheck = false;
 
     this.expectSocketClose = false;
+    this.blocked = false;
     this.freeChannels = new BitSet();
     this.channels = [{
       channel: { accept: channel0(this) },
@@ -317,6 +318,18 @@ class Connection extends EventEmitter {
   // ignored. We also have to shut down all the channels.
   toClosing (capturedStack, k) {
     var send = this.sendMethod.bind(this);
+    invalidateSend(this, 'Connection closing', capturedStack);
+
+    if(this.blocked) {
+      // if the connection is blocked, we will not receive a close-ok until the connection
+      // becomes unblocked, which may be some time. So just terminate the connection
+      // immediately
+      if (k)
+        k();
+      var s = stackCapture('Forcibly closing as connection is blocked');
+      this.toClosed(s, undefined);
+      return;
+    }
 
     this.accept = function (f) {
       if (f.id === defs.ConnectionCloseOk) {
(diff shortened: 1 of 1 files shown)


cressie176:
Hi @SimonWoolf, Thanks for the PR. I see other clients have suffer from the same challenge.

@michaelklishin, it was a while back but do you remember whether and how this issue was address with the Java client?
https://groups.google.com/g/rabbitmq-users/c/7cQXglAnWS8

Is there a defacto standard approach that amqplib should follow?

cressie176:
I've been thinking about this some more and wondering whether supplying an optional boolean "force" to the connection close method might be better. @SimonWoolf if you have time to update the PR then I'll merge. Otherwise I'll take a look, but probably not for a little while.

SimonWoolf:
> I've been thinking about this some more and wondering whether supplying an optional boolean "force" to the connection close method might be better.

to be clear, are you asking for that force arg to apply to the 'destroy() after end()' behaviour change, or to the 'close immediately if blocked' change? (or both?)

cressie176:
I was thinking of adding the option to [connection.close](https://amqp-node.github.io/amqplib/channel_api.html#model_close), then passing this through. The function also takes an optional callback which would make this a bit clumsy. Also, something I hadn't considered before, is that this wouldn't help for any non-user invoked closes.

cressie176:
Some interesting reading here...
https://rabbitmq-users.narkive.com/DSpWxIxq/rabbitmq-discuss-blocking-queue-close-connection-by-client

cressie176:
php-amqplib took a timeout approach - https://github.com/php-amqplib/php-amqplib/pull/609

kibertoad:
@SimonWoolf Ping? Are you still open to update the PR?

SimonWoolf:
> Are you still open to update the PR?

Sure, just waiting for a reply to the question I asked above (whether cressie wanted the `force` arg to close() that they requested to modify just the 'close immediately if blocked' behaviour, or also the 'end causes destroy' behaviour that I also changed)

michaelklishin:
@cressie176 we need your decision for this work to continue. Thank you.

cressie176:
Hi @SimonWoolf, @michaelklishin, @kibertoad

I'm good with the proposed fix as is, with the one exception of destroying the stream. I'm not sure whether this will have side effects.

The code base has moved on quite a bit since this PR was created. I have some time today/tomorrow, so will re-implement.

cressie176:
See https://github.com/amqp-node/amqplib/pull/845

PR #746, opened 2023-11-28
Make CI more robust

cressie176:
The GitHub test action has been having intermittent test failures. I've modified it to use netcat to check that RabbitMQ is running, which seems to have made things more stable. I also added node 20 to the matrix.


kibertoad:
@cressie176 No need to update the makefile with version for 20?

cressie176:
Good point. On it.

PR #747, opened 2023-12-13
Update channel_model.js

NevilDabre:
Got this error on the production server

(node:64) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'fields' of null at Channel.ack (/home/node/node_modules/amqplib/lib/callback_model.js:234:37) at _callee$ (/home/node/src/utils/services/rabbitmq-consumer-service.js:42:51)

My code is as follows
channel.ack(msg);

Fixed at my project but proposing the change at the library level by checking message object, thanks.

Using the same library version
  "name": "amqplib",
  "homepage": "http://amqp-node.github.io/amqplib/",
  "main": "./channel_api.js",
  "version": "0.10.3",

### DIFF
--- a/lib/channel_model.js
+++ b/lib/channel_model.js
@@ -204,6 +204,9 @@ class Channel extends BaseChannel {
   }
 
   ack(message, allUpTo) {
+    if (!message || !message.fields || typeof message.fields.deliveryTag === 'undefined') {
+      throw new Error('Invalid message parameter');
+    }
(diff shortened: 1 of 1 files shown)


cressie176:
Hi @NevilDabre
Thanks for the PR. What advantage does checking the input parameter and throwing a custom error have, over throwing the TypeError?

PR #756, opened 2024-02-22
Add connection-update-secret

cressie176:
Add support connection update-secret as per https://github.com/amqp-node/amqplib/issues/755. I have asked @mortdiggiddy to test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
-RABBITMQ_SRC_VERSION=rabbitmq_v3_2_1
+RABBITMQ_SRC_VERSION=v3.12.13
 JSON=amqp-rabbitmq-0.9.1.json
-RABBITMQ_CODEGEN=https://raw.githubusercontent.com/rabbitmq/rabbitmq-codegen
-AMQP_JSON=$(RABBITMQ_CODEGEN)/$(RABBITMQ_SRC_VERSION)/$(JSON)
+AMQP_JSON=https://raw.githubusercontent.com/rabbitmq/rabbitmq-server/$(RABBITMQ_SRC_VERSION)/deps/rabbitmq_codegen/$(JSON)
 
 NODEJS_VERSIONS='10.21' '11.15' '12.18' '13.14' '14.5' '15.8' '16.3.0' '18.1.0' '20.10.0'
 
--- a/lib/callback_model.js
+++ b/lib/callback_model.js
@@ -24,6 +24,10 @@ class CallbackModel extends EventEmitter {
     this.connection.close(cb);
   }
 
+  updateSecret(newSecret, reason, cb) {
+    this.connection._updateSecret(newSecret, reason, cb);
+  }
+
   createChannel (cb) {
     var ch = new Channel(this.connection);
     ch.open(function (err, ok) {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -361,6 +361,14 @@ class Connection extends EventEmitter {
     this.emit('close', maybeErr);
   }
 
+  _updateSecret(newSecret, reason, cb) {
+    this.sendMethod(0, defs.ConnectionUpdateSecret, {
+      newSecret,
+      reason
+    });
+    this.on('update-secret-ok', cb);
+  }
+
   // ===
   startHeartbeater () {
     if (this.heartbeat === 0)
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -1,7 +1,6 @@
-RABBITMQ_SRC_VERSION=rabbitmq_v3_2_1
+RABBITMQ_SRC_VERSION=v3.12.13
 JSON=amqp-rabbitmq-0.9.1.json
-RABBITMQ_CODEGEN=https://raw.githubusercontent.com/rabbitmq/rabbitmq-codegen
-AMQP_JSON=$(RABBITMQ_CODEGEN)/$(RABBITMQ_SRC_VERSION)/$(JSON)
+AMQP_JSON=https://raw.githubusercontent.com/rabbitmq/rabbitmq-server/$(RABBITMQ_SRC_VERSION)/deps/rabbitmq_codegen/$(JSON)
```
cressie176, line 3: I've updated to the latest spec, from the RabbitMQ monorepo.
Also closes https://github.com/amqp-node/amqplib/pull/613

lib/callback_model.js
```diff
@@ -24,6 +24,10 @@ class CallbackModel extends EventEmitter {
     this.connection.close(cb);
   }
 
+  updateSecret(newSecret, reason, cb) {
+    this.connection._updateSecret(newSecret, reason, cb);
+  }
+
```
cressie176, line 30: I was in two minds whether to default reason, but decided against it as I wasn't sure what a sensible default value would be. Happy to reconsider.

lib/connection.js
```diff
@@ -361,6 +361,14 @@ class Connection extends EventEmitter {
     this.emit('close', maybeErr);
   }
 
+  _updateSecret(newSecret, reason, cb) {
+    this.sendMethod(0, defs.ConnectionUpdateSecret, {
+      newSecret,
+      reason
+    });
+    this.on('update-secret-ok', cb);
```
cressie176: I'll change to this.once

cressie176: I used the event handler because I couldn't see another way of getting a callback into the channel0 accept function

lib/connection.js
```diff
@@ -611,6 +619,9 @@ function channel0(connection) {
     else if (f.id === defs.ConnectionUnblocked) {
       connection.emit('unblocked');
     }
+    else if (f.id === defs.ConnectionUpdateSecretOk) {
+      connection.emit('update-secret-ok');
+    }
```
cressie176, line 624: This is where channel 0 accepts the update-secret ok confirmation frame. I decided not to report an error a confirmation is received without the update-secret operation ever having been sent. 

As previously mentioned I used event emitters because I couldn't see how else to invoke the updateSecret callback.

test/callback_api.js
```diff
@@ -57,10 +57,18 @@ function waitForMessages(ch, q, k) {
...
+  });
+
 });
 
+suite('updateSecret', function() {
+  test('updateSecret', function(done) {
+    connect(kCallback(function(c) {
+      c.updateSecret(Buffer.from('new secret'), 'no reason', doneCallback(done));
+    }));
+  });
```
cressie176, line 71: Tests the callback version of updateSecret. Not a great test as there are no side effects to assert. I checked that the operation and reply are sent/received via WireShark though.

mortdiggiddy:
Will be doing some ad-hoc testing in the next few days, then I will try to add some formalized tests.

Thanks for the speedy response this is going to be helpful.

lib/connection.js
mortdiggiddy: This may already have been administered, but the new "secret" can be either a username or JWT to my knowledge.

If JWT, this would pipe into updating whatever authentication mechanism exists that decodes and verifies the new JWT.

I imagine that somewhere in the libs that a new exp (expiration) value is set which is responsible for many of the 403 access denied errors when trying to publish to a connection that has expired.

cressie176:
Great thanks. One thing to be aware of...

lib/defs.js isn't committed to GitHub, so if you were going to add the [GitHub URL](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#github-urls) to package.json, it won't work. Instead do something like...

1. fork the branch
2. run `make clean lib/defs.js` from the command line to build 
3. remove lib/defs.js from .gitignore
4. commit & push
5. Reference your fork from package.json

lib/connection.js
cressie176: The RabbitMQ docs are pretty light, but the spec only has two arguments...

```json
{
    "id": 70,
    "arguments": [
    {
      "type": "longstr",
      "name": "new-secret"
    },
    {
      "type": "shortstr",
      "name": "reason"
    }
  ],
  "name": "update-secret",
  "synchronous": true
}
```
So I assume it's also handled transparently.

PR #758, opened 2024-04-11
Commit defs

cressie176:
There are some good arguments for committing the generated lib/defs.js file

1. the github actions will run from the same lib/defs.js that the maintainers use, and that will be published
2. it will be easier for users to contribute
3. amqplib can be depended on using a github url from package.json, making it easier to test forks/branches

I've also decided to include the amqp-rabbit-0.9.1.json specification. It's not necessary, but only a small increase to the overall package size and might prove useful for someone trying to understand why an operation is missing


Makefile
```diff
@@ -9,33 +9,36 @@ _MOCHA=./node_modules/.bin/_mocha
 UGLIFY=./node_modules/.bin/uglifyjs
 NYC=./node_modules/.bin/nyc
 
-.PHONY: test test-all-nodejs all clean coverage
+.PHONY: test test-all-nodejs coverage lib/defs.js
 
-all: lib/defs.js
```
cressie176, line 14: I replaced the default "all" target with an "error" target because we don't want people running make and unintentionally generating new lib/defs.js

Makefile
```diff
@@ -9,33 +9,36 @@ _MOCHA=./node_modules/.bin/_mocha
...
+test-all-nodejs:
 	for v in $(NODEJS_VERSIONS); \
 		do echo "-- Node version $$v --"; \
 		nave use $$v $(MOCHA) -u tdd --exit -R progress test; \
 		done
 
-coverage: $(NYC) lib/defs.js
-	$(NYC) --reporter=lcov --reporter=text $(_MOCHA) -u tdd -R progress test/
+coverage: $(NYC)
+	$(NYC) --clean --reporter=lcov --reporter=text $(_MOCHA) -u tdd --exit -R progress test/
```
cressie176, line 28: Instead of the clean target deleting the .coverage directory, I thought it better to have nyc do it. 

Mocha was hanging so I added the --exit argument

Makefile
```diff
@@ -9,33 +9,36 @@ _MOCHA=./node_modules/.bin/_mocha
...
 	@echo "HTML report at file://$$(pwd)/coverage/lcov-report/index.html"
 
+lib/defs.js: clean bin/generate-defs test
```
cressie176, line 31: I wanted to be sure the new lib/defs.js were tested post generation

Makefile
kibertoad: What is causing Mocha to hang? It's generally a good idea to avoid `--exit`, because otherwise we can miss problems with resource disposal.

cressie176: I'll see if I can work it out. Most likely a connection to the broker not being closed in one of the tests.

cressie176: There are loads of test where the connection isn't being closed. Really this should be done on both success and failure, so it will take a while to fix them all. I'll create a separate issue for it and leave the --exit for now.

PR #766, opened 2024-07-08
Support for unsigned int

ebrinette:
Add support for unsigned int (uint8, uint16, uint32) as specified in https://www.rabbitmq.com/amqp-0-9-1-errata#section_3.

Ignore uint64 to avoid conflict mentionned in #646.

### DIFF
--- a/lib/codec.js
+++ b/lib/codec.js
@@ -193,6 +193,10 @@ function encodeFieldValue(buffer, value, offset) {
         tag('s');
         buffer.writeInt16BE(val, offset); offset += 2;
         break;
+    case 'unsignedshort':
+    case 'uint16':
+        tag('u');
+        buffer.writeUInt16BE(val, offset); offset += 2;
     case 'int':
     case 'int32':
         tag('I');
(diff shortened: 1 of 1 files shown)


SpiderCh:
Hello,
@cressie176, do you have any plans on merging this request and making a new release?
This issue is very painful because we're using different clients, and we have to make hacks for sending requests to services that use this library.

ebrinette:
Hi @cressie176 ,

I'm not sure anymore if there's a reason for the space. Looking at the changes it doesn't look like it. So I guess it could be changed.
I would appreciate if you had time to do it. I was thinking there might be some documentation to update as well ?

lib/codec.js
```diff
@@ -188,16 +188,30 @@ function encodeFieldValue(buffer, value, offset) {
...
+        break;
     case 'short':
     case 'int16':
         tag('s');
         buffer.writeInt16BE(val, offset); offset += 2;
         break;
+    case 'unsignedshort':
+    case 'uint16':
+        tag('u');
+        buffer.writeUInt16BE(val, offset); offset += 2;
```
cressie176, line 204: Missing a break;

PR #775, opened 2024-11-24
Codec c short support edits

cressie176:
I wanted to merge https://github.com/amqp-node/amqplib/pull/620 but couldn't because the ordering of the code type tags had changed resulting in a duplicate entry for 'u'.


cressie176:
Too many commits. Must have done something wrong.

PR #776, opened 2024-11-24
Support U tag for signed 16bit field values

cressie176:
See https://github.com/amqp-node/amqplib/pull/620 (couldn't merge because it would have created a duplicate tag for 'u')


cressie176:
Decided against this. 'U' is only valid for amqp 0.9

PR #781, opened 2025-01-27
Preserve rpc exception traceback

yotamN:


cressie176:
Hi @yotamN,

Thanks for the PR. If I understand correctly it moves the original error to the cause property of a new Error instance. 
The cause property was only added in Node: v16.9.0, and using it can actually obscure the original error

Am I missing something?

S

yotamN:
Yes, the `cause` argument was introduced in 2022, although it already appears to be widely adopted. As far as I know, Node 16 is EoL.

If you’d prefer to maintain compatibility with older Node versions, another solution would be to modify the exception’s stack trace. Let me know if you’d like me to explore that approach instead.

cressie176:
Hi @yotamN,

amqplib backwards compatibility is currently >= Node 10, so if we were to proceed, then I would want to augment the error object using something like Object.assign, however, I am still unsure what benefit wrapping the original error has. How does this preserve the RPC error stack trace?

yotamN:
I will try to show a concrete example later but the thing with `promisifiy` and in many other callback interfaces, is that they swallow the traceback. The traceback of the error inside `_rpc` will be shown up until the callback but I can't know what triggered the function from outside of the `rpc` method.

yotamN:
Just updating, I have a bit of a problem creating a complete reproduction (I got it in production but obviously it's less helpful and I can't share the code). Will keep you updated

yotamN:
Pardon my long delay, I had a busy week traveling to and from FOSDEM.
It seems that this specific case was caused by incorrect usage on my part, although there are still places where the stack traces are being swallowed, for example in the connect function itself.

    	await connect(['amqp://user:bitnami@localhost']);
    	
For example, this code won't work since it's passing an array instead of a string or an object.
 The resulting error and stack trace would look like this:
 
     Error: Handshake terminated by server: 403 (ACCESS-REFUSED) with message "ACCESS_REFUSED - Login was refused using authentication mechanism PLAIN. For details see the broker logfile."
        at afterStartOk (/tmp/tmp.67zkrnvCqW/amqplib2/lib/connection.js:169:16)
        at /tmp/tmp.67zkrnvCqW/amqplib2/lib/connection.js:104:11
        at Socket.recv (/tmp/tmp.67zkrnvCqW/amqplib2/lib/connection.js:453:9)
        at Object.onceWrapper (node:events:632:28)
        at Socket.emit (node:events:518:28)
        at emitReadable_ (node:internal/streams/readable:834:12)
        at process.processTicksAndRejections (node:internal/process/task_queues:89:21)

As you can see, there is no reference to my code.
But since this is a separate part of the code, I will open a PR once I fix this.
I think I will try to use the `cause` argument and if it fails, I will manually assign it.

PR #785, opened 2025-03-20
Feature/drop using bitsyntax pkg

ikenfin:
Drop using @acuminous/bitsyntax because it's incompatible with --disallow-code-generation-from-strings node flag, and there is no heavy bit parsing logic which cannot be replaced with some simple code


cressie176:
Hi @ikenfin,
Thank you for this. amqplib currently supports Node 0.10 onwards. Is there anything in this PR which couldn't be made compatible with this?

ikenfin:
Hi @cressie176 !
My bad. I will look how to bring backward compatibility with so legacy node and send update, thanks!

ikenfin:
@cressie176  Seems i've fixed compatibility issues, now all tests are report ok

PR #787, opened 2025-03-26
Use a frame_max of 131072 by default for RabbitMQ 4.1.0 compatibility

michaelklishin:
Just like nearly all client libraries do.

In RabbitMQ 4.1.0, the pre-authentication default changes to 8192 to accommodate for larger JWT frames, which the current value in this client won't be compatible with. See [1].

This default matches what RabbitMQ itself, the Java and .NET clients have been using since 2007.

1. https://github.com/rabbitmq/rabbitmq-server/blob/main/release-notes/4.1.0.md#initial-amqp-0-9-1-maximum-frame-size

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -66,7 +66,7 @@ function openFrames(vhost, query, credentials, extraClientProperties) {
 
     // tune-ok
     'channelMax': intOrDefault(query.channelMax, 0),
-    'frameMax': intOrDefault(query.frameMax, 0x1000),
+    'frameMax': intOrDefault(query.frameMax, 131072),
     'heartbeat': intOrDefault(query.heartbeat, 0),
 
     // open


michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/issues/13541 is the server change that bumps the minimum, first shipping in `4.1.0` in the next few weeks.

michaelklishin:
@cressie176 given that RabbitMQ `4.1.0` will ship in a few weeks, it would be great to have a new release of this library with this higher default.

With the unfortunate default used today, this library won't be able to connect to a `4.1.0` node.

cressie176:
Thanks @michaelklishin, I'll issue a new release sometime this week. Appreciate you looking out for amqplib

cressie176 (approved):

PR #801, opened 2025-09-04
feature/esm and removal of unnecessary dependencies

luas10c:
The foot is now leaner, it only has support for esm and some security fixes, such as maintenance of libs that no longer need to be used in favor of native nodejs apis.

- [x] add esm
- [x] add build with swc 
- [x] remove unnecessary dependencies

I hope everything is okay now.


cressie176:
https://github.com/amqp-node/amqplib/pull/797#issuecomment-3246009529

PR #802, opened 2025-09-04
Modernisation 3 - Enable noUnusedFunctionParameters lint rule and fix violations

cressie176:
- Enable noUnusedFunctionParameters Biome lint rule as error
- Prefix unused parameters with underscore throughout codebase
- Update package.json to increase lint diagnostics limit
- Exclude generated lib/defs.js from Biome linting
- Add project-specific linting guidelines to CLAUDE.md

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/connect.js
+++ b/lib/connect.js
@@ -47,7 +47,7 @@ function openFrames(vhost, query, credentials, extraClientProperties) {
   if (!vhost) vhost = '/';
   else vhost = QS.unescape(vhost);
 
-  var query = query || {};
+  query = query || {};
 
   function intOrDefault(val, def) {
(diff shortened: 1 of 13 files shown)


lib/connect.js
```diff
@@ -47,7 +47,7 @@ function openFrames(vhost, query, credentials, extraClientProperties) {
   if (!vhost) vhost = '/';
   else vhost = QS.unescape(vhost);
 
-  var query = query || {};
+  query = query || {};
```
cressie176, line 50: query is provided as a function parameter but was unused because it was redeclared and masked by `var`. I removed the var as the was presumably the original intent. The tests still pass.

PR #803, opened 2025-09-04
Amqplib modernisation 4 no unused variables

cressie176:
I have enabled the noUnusedVariables lint rule.
I found one variable which should have been used but was being masked (query)
Some variables I deleted, other variables I prefixed with an underscore so that the linter didn't complain


cressie176:
Merged manually into main

PR #804, opened 2025-09-04
Modernisation 5 - Replace all var declarations with let/const for modern JavaScript sta…

cressie176:
Killed 3 birds (noVar, noInnerDeclarations, noInvalidUseBeforeDeclaration) with one stone 
- Replace 674 var declarations with appropriate let/const
- Use const for immutable values (requires, constants, never-reassigned variables)
- Use let for variables that get reassigned (loop counters, mutated state)
- Modernise codebase for Node.js v18+ with proper block scoping
- Eliminate variable hoisting issues and improve code clarity
- All tests pass with zero linting errors

🤖 Generated with [Claude Code](https://claude.ai/code)


cressie176:
Merged manually into main

PR #833, opened 2026-02-13
feat: add opt-in automatic recovery for promise and callback APIs

### LINKED ISSUE #25: Support Auto-reconnection
At the moment, when a socket closes it's basically permanent. All kinds of methods are replaced in favor of functions which only `throw` and there's no reconnect support.

Even in a data center, connections can fail or time out, and reconnection would allow the connection to fail and recover from the failure, instead of simply giving up.

This might need to happen at the `connect.js` level given the current setup, but that would put messages that have been sent but not acked into an uncertain place where it's unclear if the message has reached the server or just fallen. I know from experience that trying to get reconnection to work well can be tricky. For a counterexample, see `node-amqp`'s reconnection, which appears to leak everywhere.

ShiriNmi1520:
## Summary
This PR adds a **generic, opt-in recovery mechanism** to `amqplib` for both public APIs:
- Promise API (`require('amqplib')`)
- Callback API (`require('amqplib/callback_api')`)

Recovery is enabled only when `connect(..., { recovery: ... })` is provided, so the default behaviour remains unchanged.

## Motivation
Fix #25 

## What changed
- Added shared recovery engine: `lib/recovery.js`
- Wired recovery into:
  - `channel_api.js`
  - `callback_api.js`
- Added docs in `README.md` for usage/options/events
- Added unit tests in `test/recovery.js`

## Recovery options
- `initialDelay`
- `maxDelay`
- `factor`
- `jitter`
- `maxRetries` (also accepts `maxAttempts`)
- `setup(model, client[, done])` called after each successful connect/reconnect

## Mermaid sequence chart
```mermaid
sequenceDiagram
    autonumber
    participant App
    participant Client as amqplib (recovering client)
    participant Broker as RabbitMQ

    App->>Client: connect(url, { recovery })
    Client->>Broker: open TCP + AMQP handshake
    Broker-->>Client: open-ok
    Client->>Client: run recovery.setup(...)
    Client-->>App: emit "connect"

    Note over Broker,Client: normal operation

    Broker-x Client: socket/connection closed
    Client-->>App: emit "disconnect"
    Client-->>App: emit "reconnectScheduled"

    loop retry with backoff+jitter
        Client->>Broker: reconnect + handshake
        alt success
            Broker-->>Client: open-ok
            Client->>Client: run recovery.setup(...)
            Client-->>App: emit "connect"
        else failed and retries remain
            Client-->>App: emit "connectFailed"
            Client-->>App: emit "reconnectScheduled"
        end
    end

    alt retries exhausted
        Client-->>App: emit "reconnectFailed"
    end
```

## Backward compatibility
- No behavior change unless `recovery` is explicitly configured.

## Notes / limitations
- This provides connection-level recovery and setup replay.
- It does not guarantee transparent replay of in-flight channel operations across disconnects.


lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+    };
+  }
+
+  _unbindModel() {
+    if (!this._modelListeners) return;
+    const { model, onClose, onBlocked, onUnblocked, onConnectionError } = this._modelListeners;
+    model.removeListener('close', onClose);
+    model.removeListener('blocked', onBlocked);
+    model.removeListener('unblocked', onUnblocked);
+    model.removeListener('error', onConnectionError);
```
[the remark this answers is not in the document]
ShiriNmi1520: No worries, take your time.
I'll implement modifications once reviewed.

cressie176:
@ShiriNmi1520, thank you for your patience. It is a really good PR. Just have a lot on at the moment, so taking time to get to it.

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+  return Number.isFinite(n) ? n : fallback;
+}
+
+function normaliseRecoveryOptions(recovery) {
+  const source = recovery === true ? {} : recovery || {};
+  const initialDelay = Math.max(0, toFiniteNumber(source.initialDelay, DEFAULT_RECOVERY.initialDelay));
+  const maxDelay = Math.max(initialDelay, toFiniteNumber(source.maxDelay, DEFAULT_RECOVERY.maxDelay));
+  const factor = Math.max(1, toFiniteNumber(source.factor, DEFAULT_RECOVERY.factor));
+  const jitter = Math.min(1, Math.max(0, toFiniteNumber(source.jitter, DEFAULT_RECOVERY.jitter)));
+  const maxRetries = source.maxRetries === undefined ? source.maxAttempts : source.maxRetries;
```
cressie176: Having both maxAttempts and maxRetries is slightly ambiguous. Strictly speaking an "attempt" would include the first connection, whereas as "retry" does not include the first connection.

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+
+    const attempt = this._attempt + 1;
+    const delay = calculateDelay(this._recovery, attempt);
+    this._attempt = attempt;
+
+    this.emit('reconnectScheduled', { attempt, delay, error: err });
+    this._timer = setTimeout(() => {
+      this._timer = null;
+      this._connect();
+    }, delay);
```
cressie176, line 301: Given there could be a reasonable delay between retries, it is probably worth calling `setTimeout(...).unref()`. Otherwise the timer could block graceful shutdown.

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+    };
+
+    const onBlocked = (reason) => this.emit('blocked', reason);
+    const onUnblocked = () => this.emit('unblocked');
+    const onConnectionError = (err) => this.emit('connectionError', err);
+
+    model.on('close', onClose);
+    model.on('blocked', onBlocked);
+    model.on('unblocked', onUnblocked);
+    model.on('error', onConnectionError);
```
cressie176, line 227: Is this missing the `update-secret-ok` event?

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+  updateSecret: callbackUpdateSecret,
+};
+
+function wireCommonEvents(source, target) {
+  ['connect', 'disconnect', 'connectFailed', 'reconnectScheduled', 'reconnectFailed', 'blocked', 'unblocked', 'connectionError'].forEach(
+    (eventName) => {
+      source.on(eventName, (...args) => target.emit(eventName, ...args));
+    },
+  );
+}
```
cressie176, line 372: Is this missing the `update-secret-ok` event?

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+      this._scheduleReconnect(error);
+    } finally {
+      this._connecting = false;
+    }
+  }
+
+  _scheduleReconnect(err) {
+    if (this._stopped) {
+      this._rejectInitialConnection(err);
+      this._rejectPendingWaiters(err);
```
cressie176: I believe these will have already been called on lines 148-149

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+class RecoveringPromiseModel extends EventEmitter {
+  constructor(openModel, recovery) {
+    super();
+    this._core = new RecoveringCore(openModel, recovery, PROMISE_ADAPTERS);
+    wireCommonEvents(this._core, this);
+  }
+
+  waitForConnect() {
+    return this._core.waitForConnect();
+  }
```
cressie176, line 384: If I have followed correctly, RecoveringPromiseModel.waitForConnect will yield the amqplib model. Is this what you intended? I would have thought it should yield the instance of the RecoveringPromiseModel

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+  constructor(openModel, recovery) {
+    super();
+    this._core = new RecoveringCore(openModel, recovery, CALLBACK_ADAPTERS);
+    wireCommonEvents(this._core, this);
+  }
+
+  waitForConnect(cb) {
+    makeCallback(this._core.waitForConnect(), cb);
+    return this;
+  }
```
cressie176, line 417: If I have followed correctly, RecoveringCallbackModel.waitForConnect will yield the amqplib model. Is this what you intended? I would have thought it should yield the instance of the RecoveringCallbackModel

lib/recovery.js
```diff
@@ -0,0 +1,457 @@
...
+      settled = true;
+      if (err) reject(toError(err, 'Recovery setup failed'));
+      else resolve();
+    }
+
+    try {
+      if (setup.length >= 3) {
+        setup(model, client, done);
+      } else {
+        Promise.resolve(setup(model, client)).then(() => done(), done);
```
cressie176: Why pass the client (RecoveringCore) to the setup function?
Should this not be kept internal to prevent accidental misuse?

cressie176 (changes requested):

cressie176:
@ShiriNmi1520 I've finished my review now. I have a few questions which I suspect will lead to some changes. I also wanted to add that this is some of most well written code I have ever seen. Very well crafted. Thank you.

ShiriNmi1520:
> @ShiriNmi1520 I've finished my review now. I have a few questions which I suspect will lead to some changes. I also wanted to add that this is some of most well written code I have ever seen. Very well crafted. Thank you.

Appreciate your review, please allow me some time to implement modifications based on your review :)

lib/recovery.js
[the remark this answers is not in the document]
ShiriNmi1520: Replaced the bare `removeListener('error', …)` with a follow-up `model.on('error', () => {})` so any late errors emitted while the closed connection tears down get swallowed instead of bubbling up unhandled. Fixed in 42c5962.

lib/recovery.js
ShiriNmi1520: Agreed, the dual key is ambiguous. Dropped the undocumented `maxAttempts` alias entirely — the implemented semantics already match “retries excluding the initial connect,” so `maxRetries` is the only public key now. Fixed in 7c39e80.

lib/recovery.js
ShiriNmi1520: Marked the reconnect `setTimeout` as unref’d so a pending retry no longer keeps the event loop alive. Fixed in 42c5962.

lib/recovery.js
ShiriNmi1520: Added missing `update-secret-ok` to the forwarded event list so it propagates from the core through to the outer wrapper. Fixed in 2a37721

lib/recovery.js
ShiriNmi1520: Added missing `update-secret-ok` to the forwarded event list so it propagates from the core through to the outer wrapper. Fixed in 2a37721

lib/recovery.js
ShiriNmi1520: Confirmed redundant. The only path that hits this branch is `_connect()`’s catch firing after `close()` already rejected `_initialWait` and the pending waiters; both reject helpers are idempotent at that point, so the calls were no-ops. Removed; the `if (this._stopped) return;` is sufficient. Fixed in 86d3b22.

lib/recovery.js
ShiriNmi1520: That was a bug, not an intent — yielding the raw amqplib model meant that callers could hold a reference that would become invalid after a reconnect, fully bypassing recovery. `RecoveringPromiseModel.waitForConnect` now resolves with `this`, and the internal `_initialWait` no longer carries the model. Also dropped the now-redundant `.then(() => recovering)` shim in `connectWithRecoveryPromise`. Fixed in 8331773.

lib/recovery.js
ShiriNmi1520: Same fix applied to the callback path: `RecoveringCallbackModel.waitForConnect(cb)` now invokes the callback with `(null, this)`. `connectWithRecoveryCallback` simplified accordingly. Fixed in 8331773 (same commit).

lib/recovery.js
ShiriNmi1520: Agreed, it shouldn’t be exposed — `RecoveringCore` is internal and leaking it into `setup` puts private methods (`_connect`, `_scheduleReconnect`, etc.) on the public surface. Dropped the parameter; `setup` is now `(model)` for the promise variant or `(model, done)` for the
callback variant. README updated to match. Fixed in 7c39e80.
If you’d prefer to keep a recovery-aware handle reachable from `setup`, the cleanest alternative would be passing the outer `RecoveringPromiseModel` / `RecoveringCallbackModel` wrapper instead of the core — happy to switch to that if you’d rather not narrow the
signature.

lib/recovery.js
cressie176: Hi @ShiriNmi1520,
noticed a couple more things (sorry). The event name pattern so far uses `-` instead of camel case. Can you also changed `reconnectScheduled` to `reconnect-scheduled` and so on please?

ShiriNmi1520:
Thank you again for taking the time to review everything so carefully. Each of your points was helpful and made a difference. I've pushed five follow-up commits, with each one addressing a separate change.

lib/recovery.js
```diff
@@ -0,0 +1,467 @@
...
+      if (this._stopped) return;
+
+      const err = toError(maybeErr, 'Connection closed');
+      this.emit('disconnect', err);
+      this._scheduleReconnect(err);
+    };
+
+    const onBlocked = (reason) => this.emit('blocked', reason);
+    const onUnblocked = () => this.emit('unblocked');
+    const onConnectionError = (err) => this.emit('connectionError', err);
```
cressie176: Does this mean amqplib model "error" events are translated to RecoveringXModel "connectionError" events?

If so, this would mean that any existing error handlers users had registered would no longer work. Furthermore, if the user was relying on the "error" event to crash their application, this would stop working too.

lib/recovery.js
```diff
@@ -0,0 +1,467 @@
...
+    'blocked',
+    'unblocked',
+    'connectionError',
+    'update-secret-ok',
+  ].forEach((eventName) => {
+    source.on(eventName, (...args) => target.emit(eventName, ...args));
+  });
+}
+
+class RecoveringPromiseModel extends EventEmitter {
```
cressie176, line 374: If I have followed correctly, RecoveringPromiseModel.waitForConnect will yield the amqplib model. Is this what you intended? I would have thought it should yield the instance of the RecoveringPromiseModel

lib/recovery.js
ShiriNmi1520: Renamed `connectFailed` / `reconnectScheduled` / `reconnectFailed` to their kebab-case forms in 8ca2023. The `connectionError` rename is in the same commit but driven by your other comment — see [below](https://github.com/amqp-node/amqplib/pull/833#discussion_r3200109989).

lib/recovery.js
ShiriNmi1520: You're right — translating model `error` → wrapper `connectionError` would silently break the existing `connection.on('error', …)` handlers and lose the EventEmitter crash-on-unhandled-error safety net.

Renamed the wrapper event to `error` in 8ca2023. Now, any handler the user previously had on the underlying connection keeps working when wrapped in recovery, and an unhandled error on the wrapper crashes the process the same way it would have on the underlying
connection.

Note: the internal `model.on('error', () => {})` no-op listener we added during `_unbindModel` (commit 42c5962) is still attached only to the closed/stale model after we've already handled disconnect — it's there to swallow late teardown errors on a connection we no longer care about, not to silence errors on the active one.

lib/recovery.js
ShiriNmi1520: I think this thread may have been reopened against the new line position — the fix you originally asked for is already in 8331773.

cressie176 (approved):

cressie176:
My bad re unref - adding it broke the recovery tests in Node 18, 20 and 22. Thinking about it, we didn't want to unref the timer between recovery attempts, as if the event loop was drained it would have caused the application to exit. You already clear the timer when close is called, so the user can still exit the application gracefully.

cressie176:
Published as amqplib v1.1.0. Thank you very much

PR #857, opened 2026-07-10
Add custom reconnect delay strategy option

### LINKED ISSUE #855: recovery: allow a custom delay strategy (calculateDelay option)
The built-in recovery computes reconnect delays in `lib/recovery.js` with a fixed strategy: an exponential base capped at `maxDelay`, plus symmetric jitter around the base (`offset = Math.random() * jitterPart * 2 - jitterPart`). Two consequences for consumers:

- the jitter shape is fixed — full jitter (`uniform[0, cap]`), decorrelated jitter, or fixed-step schedules are not expressible;
- the cap applies to the base *before* jitter, so the effective delay can exceed `maxDelay` by up to the jitter fraction (with `jitter: 1`, up to 2× `maxDelay`).

**Proposal:** accept an optional function in the recovery options, following the precedent of the functional `setup` option:

```js
await amqplib.connect(url, {
  recovery: {
    // ms; attempt starts at 1; falls back to the built-in strategy when absent
    calculateDelay: (attempt) => number,
  },
});
```

When absent, behavior is unchanged. `normaliseRecoveryOptions` currently drops unknown keys, so the addition is backward-compatible.

**Context:** we maintain an AMQP adapter built on the v2 recovery (Connectum framework). Some strategies can be approximated with the numeric knobs (e.g. full jitter via `jitter: 1` plus halved `initialDelay`/`maxDelay`), but that is fragile against internal changes and does not cover step schedules or other shapes. A supported hook would remove the need to own a separate reconnect loop just for delay control.

Happy to send a PR if the direction is acceptable.

GiHoon1123:
## Summary

Closes #855.

Adds an optional `calculateDelay(attempt)` recovery setting for custom reconnect delays such as full jitter, decorrelated jitter, or a fixed schedule.

```js
const connection = await amqplib.connect('amqp://localhost', {
  recovery: {
    maxRetries: Infinity,
    calculateDelay(attempt) {
      return Math.min(30000, 100 * 2 ** (attempt - 1));
    },
  },
});
```

The option works with both the promise and callback APIs.

Also caps the built-in jittered delay at `maxDelay`. The base is capped at `maxDelay / (1 + jitter)` rather than `maxDelay` itself, so the max jitter offset lands exactly on the cap instead of every delay above it collapsing onto the same value.

## Behavior

- `calculateDelay` errors (throw, or an invalid return value) now propagate synchronously instead of falling back silently.
- `maxDelay` only bounds the built-in strategy - `calculateDelay` callers cap their own delay.
- Return values must be finite, non-negative numbers; `'1000'` isn't coerced.
- Without `calculateDelay`, the built-in strategy is used.

## Tests

`node --test test/recovery.test.js` (12 passing), `npm run lint`, `npm run typecheck` pass. Also ran `npm test` (326 passing) against a local RabbitMQ broker.


intech:
Thanks for picking this up, @GiHoon1123 — this is exactly the shape I proposed in #855. I maintain an AMQP adapter (Connectum) built on the v2 recovery, which was the original motivation for the request, so a review from the requester's side:

**The API matches the proposal.** `calculateDelay(attempt)` is opt-in and 1-based (consistent with the built-in `factor ** (attempt - 1)`), absence keeps the current behaviour, and `normaliseRecoveryOptions` accepting it only when it is a function is the right guard. The validation predicate — `typeof === 'number' && Number.isFinite(...) && >= 0`, no coercion, so `'1000'` is rejected rather than silently accepted — is the correct strictness, and falling back to the built-in strategy instead of throwing out of an unawaited reconnect callback is the right call for a path that would otherwise take the process down.

Two things:

**1. CI has not run.** The PR currently reports zero check runs, which looks like the pending-approval gate for a first-time contributor rather than anything wrong with the branch. @cressie176 — could you approve the workflow run so the suite reports? Nothing else here is blocked on the author.

**2. Scope of "Closes #855".** #855 raised two points, and this PR addresses one of them. The hook lets a caller *avoid* the cap overshoot by owning the number, but `calculateBuiltinDelay` is unchanged, so the numeric-knob path still applies `maxDelay` to the base *before* jitter and can exceed the cap by the jitter fraction (up to ~2x at `jitter: 1`). Worth either fixing that here as well, or keeping a tracking issue open so the automatic close of #855 does not drop it.

Happy to help validate: the PR notes the recovery tests use an in-memory model, and we have a testcontainers-based RabbitMQ recovery harness. I can run this branch against a live broker — measuring the actual inter-attempt intervals across real connection drops, plus the fallback-on-throw and fallback-on-invalid paths — and report back, or contribute a broker-level integration test in the style of the existing suite, whichever is more useful to you.

GiHoon1123:
Good catch. I pushed a follow-up that clamps the built-in delay after jitter and added a deterministic test for the maximum positive jitter case.

The recovery tests, lint, and typecheck pass locally.

intech:
The clamp looks right, and thanks for turning it around so fast — that closes the second half of #855.

One observation about the shape it produces, worth considering before this lands. Clamping after symmetric jitter puts a point mass exactly on the cap: at steady state `base` has already saturated to `maxDelay`, so `offset > 0` — half the draws at any `jitter` — now collapses onto `maxDelay` itself. With the default `jitter: 0.2` that means roughly half of all reconnects in a fleet fire at the same instant, which is the opposite of what the jitter is there for. Before the change those draws were spread over `(maxDelay, maxDelay × 1.2]`.

The usual ways to keep the cap without the atom are to shrink the base (`min(maxDelay / (1 + jitter), ...)`) or to jitter downward only (`base × (1 - jitter * Math.random())`). Both shift the mean, so it is a real trade-off rather than an obvious win — your call, and the maintainer's. Flagging it because it changes the built-in distribution for every existing user, not only for those who were exceeding the cap.

Either way `calculateDelay` makes it a non-issue for anyone who cares about the exact shape, which is an argument for landing both in the same release.

cressie176:
Thanks for the PR @GiHoon1123 and the thoughtful comments @intech. I plan to go through this tonight.

lib/recovery.js
```diff
@@ -65,11 +67,33 @@ function toError(err, fallbackMessage) {
...
+function resolveDelay(core, recovery, attempt) {
+  if (recovery.calculateDelay) {
+    try {
+      const custom = recovery.calculateDelay(attempt);
+      if (typeof custom === 'number' && Number.isFinite(custom) && custom >= 0) {
+        return Math.round(custom);
+      }
+      throw new Error(`calculateDelay must return a finite, non-negative number of milliseconds (got ${String(custom)})`);
+    } catch (err) {
+      setImmediate(() => core.emit('handler-error', toError(err, 'calculateDelay failed'), 'calculateDelay'));
```
cressie176: handler-error is for when a custom error handler throws an exception. I don't think we should use it in this instance. I think I would prefer throwing the error synchronously

lib/recovery.js
```diff
@@ -65,11 +67,33 @@ function toError(err, fallbackMessage) {
...
+  return Math.min(recovery.maxDelay, Math.max(0, Math.round(base + offset)));
+}
+
+// A broken custom `calculateDelay` (throws, or returns something other than
+// a finite, non-negative number) must not stop reconnection: this runs from
+// an unawaited setTimeout/reconnect callback, so throwing here would surface
+// as an uncaught exception or unhandled rejection and crash the process.
+// Fall back to the built-in strategy and report the problem via
+// `handler-error` unconditionally - a harmless no-op if nobody's listening.
+function resolveDelay(core, recovery, attempt) {
```
cressie176: "resolve" in "resolveDelay" implies asynchronous calculation. I might just call it "calculateDelay" and have it delegate to the supplied function or built in.

lib/recovery.js
```diff
@@ -65,11 +67,33 @@ function toError(err, fallbackMessage) {
...
+  if (recovery.calculateDelay) {
```
cressie176: Prefer guard conditions (immediate return) rather than long lived if and nesting

README.md
```diff
@@ -178,6 +178,33 @@ amqplib.connect(
 
 Without `recovery` options, behavior is unchanged.
 
+### Custom delay strategy
```
cressie176, line 181: Needs to be clear that the user must implement maxDelay otherwise they might assume the library will enforce it

lib/recovery.js
```diff
@@ -65,11 +67,33 @@ function toError(err, fallbackMessage) {
   return new Error(err ? String(err) : fallbackMessage);
 }
 
-function calculateDelay(recovery, attempt) {
+function calculateBuiltinDelay(recovery, attempt) {
   const base = Math.min(recovery.maxDelay, recovery.initialDelay * recovery.factor ** (attempt - 1));
   const jitterPart = base * recovery.jitter;
   const offset = jitterPart > 0 ? Math.random() * jitterPart * 2 - jitterPart : 0;
-  return Math.max(0, Math.round(base + offset));
+  return Math.min(recovery.maxDelay, Math.max(0, Math.round(base + offset)));
```
cressie176: As @intech pointed out, enforcing maxDelay means the jitter will not be applied once maxDelay has been exceeded. I would prefer to keep the jitter, or capping the base to `maxDelay - maxJitter` so there is always room for the jitter?

cressie176 (changes requested):

GiHoon1123:
Thanks for the review - pushed changes addressing all of this:

- Renamed `resolveDelay` to `calculateDelay`, switched to a guard clause.
- `calculateDelay` errors now propagate synchronously instead of falling back via `handler-error`.
- README now says `maxDelay` only bounds the built-in strategy.
- Fixed the jitter point mass @intech flagged - base is capped at `maxDelay / (1 + jitter)` now instead of `maxDelay`. Added a regression test for it.

One consequence of the synchronous throw: it happens before `_attempt`/`_timer` get updated in `_scheduleReconnect`, so a broken `calculateDelay` doesn't just fail one attempt - the connection stops reconnecting entirely after that, with no `reconnect-failed` or `error` event. Flagging it in case it's worth addressing here or separately.

